PR #17, opened 2019-12-05
Code coverage for MappingConfig

kobylynskyi:
Relates to https://github.com/kobylynskyi/graphql-java-codegen/pull/16


kobylynskyi:
@valinha please review

kobylynskyi:
@valinha you are right. Lombok annotations do not work for generated sources.
I removed all the stuff related to this and kept only MappingConfig test (for the class which you wrote yesterday).

Issue #14, opened 2019-11-23
Subscription methods with wrong return type


Smolli:
Having a subscription in the schema generates a Subscription class with methods that have the wrong return type. Subscriptions have to return `org.reactivestreams.Publisher<T>`.

Given schema:

```graphql
schema {
  subscription: Subscription
}
type Subscription {
 projectsSub: Project!
}
type Project {
  id: String!
  name: String!
}
```

Generated class:
```java
public interface ProjectsSubSubscription {
    @javax.validation.constraints.NotNull
    Project projectsSub() throws Exception;
}
```

Expected:
```java
public interface ProjectsSubSubscription {
    @javax.validation.constraints.NotNull
    org.reactivestreams.Publisher<Project> projectsSub() throws Exception;
}
```

kobylynskyi:
@Smolli please provide more info:
* Your schema
* Actually generates class
* What’s expected

Smolli:
Hi @kobylynskyi,

I updated my description

kobylynskyi:
@Smolli, done in https://github.com/kobylynskyi/graphql-java-codegen/releases/tag/v1.3.0
Will release maven and gradle plugins later today.

PR #18, opened 2019-12-11
Subscription methods with custom return type #14

kobylynskyi:
@Smolli, please review


Smolli:
Hi @kobylynskyi, looks fine to me.

Issue #27, opened 2020-03-30
Union type members don't implement union interface if declared in separate file


joffrey-bion:
**Describe the bug**
If the types that are part of a union type are declared in other files than the union itself, they don't implement the union interface in the generated code.

**To Reproduce**

`schema1.graphqls`:
```
type A {
    fieldA: String
}
```
`schema2.graphqls`:
```
type B {
    fieldB: String
}
union U = A | B
```
This will generate `class B implements U` but a `class A` that does not implement `U`.

**Expected behavior**
All members of the union type should implement the interface.

I believe this has to do with the schema files being processed independently instead of reading all schema files to populate a common registry and then generating code from the common registry.

kobylynskyi:
@joffrey-bion I can't reproduce the issue.

schema:
`union Assignee = Bot | Mannequin | Organization | User`

Generated classes:
`public class Bot implements Assignee`
`public class Mannequin implements Assignee`
`public class Organization implements Assignee`
`public class User implements Assignee`

joffrey-bion:
Mmmh interesting. Maybe it has to do with the order of the files, then? I don't know how that could be though.
I'll try to reproduce again.

joffrey-bion:
I added a test case for multi-file unions and it does reproduce the problem.

Please take a look at the pull request https://github.com/kobylynskyi/graphql-java-codegen/pull/28 for the fix.

kobylynskyi:
@joffrey-bion Thank you! Will take a look within an hour

PR #28, opened 2020-03-30
Parse all schemas as one Document

joffrey-bion:
This makes use of the MultiSourceReader provided by the graphQL library.
Using this reader allows to read a single document from multiple files,
while correctly associating line numbers and errors to their original
source file.

This doesn't bring any overhead because the method Parser.parseDocument(String) that was previously used already created a MultiSourceReader itself.
Note that the previous method didn't associate the source file name with the content used for parsing, while the new one does.

It allows to solve cross-reference problems with union types for instance.

Resolves:
#27

### DIFF
--- a/src/main/java/com/kobylynskyi/graphql/codegen/GraphqlDocumentParser.java
+++ b/src/main/java/com/kobylynskyi/graphql/codegen/GraphqlDocumentParser.java
@@ -2,17 +2,33 @@ package com.kobylynskyi.graphql.codegen;
 
 import com.kobylynskyi.graphql.codegen.utils.Utils;
 import graphql.language.Document;
+import graphql.parser.MultiSourceReader;
 import graphql.parser.Parser;
 
 import java.io.IOException;
+import java.util.Collections;
+import java.util.List;
 
 class GraphqlDocumentParser {
 
     private static final Parser GRAPHQL_PARSER = new Parser();
 
     static Document getDocument(String schemaFilePath) throws IOException {
-        String fileContent = Utils.getFileContent(schemaFilePath);
-        return GRAPHQL_PARSER.parseDocument(fileContent);
+        return getDocument(Collections.singletonList(schemaFilePath));
     }
 
+    static Document getDocument(List<String> schemaPaths) throws IOException {
+        MultiSourceReader reader = createMultiSourceReader(schemaPaths);
+        return GRAPHQL_PARSER.parseDocument(reader);
+    }
+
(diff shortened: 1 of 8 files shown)


kobylynskyi:
@joffrey-bion some tests are failing. Could you please take a look?

src/main/java/com/kobylynskyi/graphql/codegen/GraphqlDocumentParser.java
```diff
@@ -2,17 +2,33 @@
...
+    static Document getDocument(List<String> schemaPaths) throws IOException {
+        MultiSourceReader reader = createMultiSourceReader(schemaPaths);
+        return GRAPHQL_PARSER.parseDocument(reader);
+    }
+
+    private static MultiSourceReader createMultiSourceReader(List<String> schemaPaths) throws IOException {
+        MultiSourceReader.Builder builder = MultiSourceReader.newMultiSourceReader();
+        for (String path : schemaPaths) {
+            // appending EOL to ensure that schema tokens are not mixed in case files are not properly ended with EOL
+            String content = Utils.getFileContent(path) + "\n";
```
kobylynskyi: Please use `System.lineSeparator()` instead of `"\n"`

joffrey-bion:
Yes, I'm fixing the tests.
It's because of the whitespace at the end of some test files.

joffrey-bion:
I've made a separate PR, please take a look: https://github.com/kobylynskyi/graphql-java-codegen/pull/29

This one will be based on the 29, so the diff will get cleaner when the other one is merged.

kobylynskyi:
Awesome, thank you. #29 was merged.
Will merge this one after the following change will be done:
https://github.com/kobylynskyi/graphql-java-codegen/pull/28#discussion_r400193578

src/main/java/com/kobylynskyi/graphql/codegen/GraphqlDocumentParser.java
joffrey-bion: Good spot, will make this change.

This shouldn't matter much since the built string is simply fed to the GraphQL parser (we're not generating a file out of it), but maybe the parser does use platform-specific line endings, who knows 😄

joffrey-bion:
I added the requested change. Please don't hesitate to add any other comment.

kobylynskyi:
@joffrey-bion thank you so much for the contribution.
Will release the change in 1.4.2 today

joffrey-bion:
@kobylynskyi Well, thanks to you for being so quick to review, merge and release!

PR #29, opened 2020-03-30
Cleanup: remove residual trim() added in previous PR

joffrey-bion:
In commit 96f7aa356cd17d2ea7f20b50235fe2eb492cca36, I had added a `trim()` call in `Utils.getFileContent()` because during my tests the generated files didn't contain an EOL at the end of the file, but some of my test files did.

I didn't realize this function was called in production code at the time.

This PR reverts this change but also adds a utility function for the tests to enable comparing trimmed file contents.


kobylynskyi:
Will release the change in 1.4.2. 
@joffrey-bion Thanks for your contribution!

PR #31, opened 2020-03-31
Support fields resolvers #25

### LINKED ISSUE #25: Support parameterized fields resolvers
In the GraphQL specification, some fields may have arguments:
https://graphql.org/learn/queries/#arguments

This schema for instance:
```
type Query {
    users: [Person]!
}
type Person {
    name: String!
    height(unit: Unit): Int!
}
enum Unit {
    METERS, FEET
}
```

Currently, `graphql-java-codegen` generates a `Person` class with both `name` and `height` fields. However, the height field is parameterized and should be defined in a separate resolver (with the parameters as method arguments).

A solution that would match the https://github.com/graphql-java-kickstart/graphql-java-tools approach would be to avoid generating the `height` field as a field/getter/setter on the `Person` type, but instead generate a `PersonResolver` interface with a method to resolve that field, as [defined by the GraphQL Java Tools project](https://www.graphql-java-kickstart.com/tools/schema-definition/#resolvers-and-data-classes).

The method should look like:
```java
import graphql.schema.DataFetchingEnvironment;

interface PersonResolver {
    public Integer height(Person person, Unit unit, DataFetchingEnvironment env);
}
```

Alternatively, or in the meantime, an option to exclude parameterized fields from the generated classes would be a start.

kobylynskyi:
Resolves #25

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,24 +13,26 @@ Please refer to:
 
 ## Supported Options
 
-| Key                       | Data Type          | Default value                             | Description |
-| ------------------------- | ------------------ | ----------------------------------------- | ----------- |
-| graphqlSchemaPaths        | List(String)       | None                                      | GraphQL schema locations. You can supply multiple paths to GraphQL schemas. |
-| packageName               | String             | Empty                                     | Java package for generated classes. |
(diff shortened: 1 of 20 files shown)


joffrey-bion:
This looks absolutely awesome!

Issue #30, opened 2020-03-31
Allow a directory of schemas to be passed in


joffrey-bion:
**Is your feature request related to a problem? Please describe.**
When we want to break the schema down into many parts (for scalability), it's very inconvenient to have to maintain a list of schemas to process (either in the code or in a gradle/maven plugin).

**Describe the solution you'd like**
It would be nice to be able to specify a directory so that `graphql-java-codegen` processes all the schema files inside.

A couple of useful related options to consider:
- `recursive`: whether directories should be explored recursively to find schemas
- `schemaFilePattern`: a pattern ([Java style](https://docs.oracle.com/javase/7/docs/api/java/util/regex/Pattern.html)) to match the schema files to process (would most likely default to `.*\.graphqls`)
- `excludeFiles`: a list of schema files to exclude

**Describe alternatives you've considered**
The alternatives so far are:
- to list all files manually (most likely in the maven plugin)
- to write custom code (maybe as another plugin/task) to build the list of files

kobylynskyi:
This would be a very useful feature to have!
I hope that I will have time for this one late this week or early-next week (as you know right now I am working on #25 which is quite big :smile:)

joffrey-bion:
Maybe I can contribute this one if you're ok, it should be fairly simple.

kobylynskyi:
@joffrey-bion sure, I would appreciate it if you can work on this feature.
From the implementation POV, I like your option to remove schemas from the constructor and pass them as a parameter to generate() method. We are free to change that because as of now there are only 2 usages of this class (gradle and maven plugins)

kobylynskyi:
@joffrey-bion 
I would like to reopen this feature request in order to track the changes which need to be made in gradle and maven plugins.

joffrey-bion:
Thanks for reopening, let's keep track of the gradle plugin updates here.

PR #32, opened 2020-03-31
Add SchemaFinder to enable directory support in maven/gradle plugins

joffrey-bion:
This didn't require any change in the current `GraphqlCodegen` implementation after all.
We can actually just use the `SchemaFinder` from the caller code (plugin) to get a list of schemas, and then call the generator with that list.

Resolves:
https://github.com/kobylynskyi/graphql-java-codegen/issues/30

### DIFF
--- a/src/main/java/com/kobylynskyi/graphql/codegen/supplier/SchemaFinder.java
+++ b/src/main/java/com/kobylynskyi/graphql/codegen/supplier/SchemaFinder.java
@@ -1,0 +1,95 @@
+package com.kobylynskyi.graphql.codegen.supplier;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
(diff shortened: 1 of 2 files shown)


joffrey-bion:
I'm done with the tests, there was an issue about file traversal order. On the CI the order was not the same. 
I realized that the depth-first order of the spec didn't guarantee anything about same-level order.

I fixed the tests by sorting the returned schemas alphabetically

joffrey-bion:
Thanks a lot for merging! I'll make changes to the plugins once a new version of the core lib is released

kobylynskyi:
I will try to work on #33 first if you don't mind. So that you can make changes in the same repo.

joffrey-bion:
@kobylynskyi sure that's totally fine. It's really nice that you're doing this. Thanks a lot for your reactivity and availability.

PR #34, opened 2020-04-01
Gather library and maven/gradle plugins in a single repo

### LINKED ISSUE #33: Gather plugins and core library in same repo?
Having different projects/artifacts for the Maven/Gradle plugins and the core library makes total sense. However, they don't need to be in separate GitHub repositories.

Cross-project changes and releases would be easier if they were in the same git repo:
- we could make changes to both the plugins and the library in the same commit, always keeping compatibility
- we wouldn't need to wait for the library release to make changes to the Maven/Gradle plugins
- we could test the plugins against library changes in the CI of the whole multi-project build

What are your thoughts on this?

kobylynskyi:
Closes #33


kobylynskyi:
@joffrey-bion could you please take a look at this?

plugins/graphql-java-codegen-gradle-plugin/graphql-codegen-gradle-plugin/build.gradle
```diff
@@ -0,0 +1,48 @@
...
+apply plugin: "maven-publish"
+
+group = "io.github.kobylynskyi"
+version = "1.4.2"
+description = "Provides a task for generating Java code based on GraphQL schema"
+
+dependencies {
+    compile gradleApi()
+
+    compile 'io.github.kobylynskyi:graphql-java-codegen:1.4.2'
```
joffrey-bion, line 26: This still means that the library's artifact needs to be published (unless I missed something).
We could use an actual multi-project gradle build (and declare dependencies like `compile project(":graphql-java-codegen")`.
However, I'm not sure about the maven one. I don't know how to setup a Gradle build for the maven plugin subproject

kobylynskyi: You are right.
After some research, I was not able to find a good solution for importing a Maven project inside a Gradle project or building a maven plugin inside Gradle. We can leave this as an open question and make an improvement later.

But for right now I've added a [Development process](https://github.com/kobylynskyi/graphql-java-codegen/blob/89df7a3332e99011d7f953ffc4d0fc495193e586/CONTRIBUTING.md#development-process) instructions which explain how to do the changes the plugins without publishing the core library to maven central.
@joffrey-bion Please let me know what you think. Thanks!

CONTRIBUTING.md
```diff
@@ -0,0 +1,53 @@
...
+   ```shell script
+   # This will install the library (including your recent changes) in your local maven repository.
+   ./gradlew clean install
+   ```
+   
+4. Build the plugin project with updated `graphql-java-codegen` library.
+
+   ```shell script
+   # Build Gradle plugin
+   cd plugins/graphql-java-codegen-gradle-plugin/graphql-codegen-gradle-plugin
```
joffrey-bion, line 32: Is this correct or is there an extra `/graphql-codegen-gradle-plugin` here?

plugins/graphql-java-codegen-gradle-plugin/graphql-codegen-gradle-plugin/build.gradle
```diff
@@ -0,0 +1,48 @@
...
+apply plugin: "maven-publish"
+
+group = "io.github.kobylynskyi"
+version = "1.5.0-SNAPSHOT"
+description = "Provides a task for generating Java code based on GraphQL schema"
+
+dependencies {
+    compile gradleApi()
+
+    compile 'io.github.kobylynskyi:graphql-java-codegen:1.5.0-SNAPSHOT'
```
joffrey-bion, line 26: This should use the current version to work, right? Maybe it should be a variable then, because otherwise we'll need to change the version in too many places when a release is made.

plugins/graphql-java-codegen-maven-plugin/pom.xml
```diff
@@ -0,0 +1,247 @@
...
+        <version.maven-resources-plugin>3.1.0</version.maven-resources-plugin>
+        <version.maven-source-plugin>3.1.0</version.maven-source-plugin>
+        <version.maven-javadoc-plugin>3.1.1</version.maven-javadoc-plugin>
+        <version.maven-deploy-plugin>3.0.0-M1</version.maven-deploy-plugin>
+        <version.nexus-staging-maven-plugin>1.6.8</version.nexus-staging-maven-plugin>
+        <version.maven-release-plugin>2.5.3</version.maven-release-plugin>
+        <version.maven-scm-provider-gitexe>1.11.2</version.maven-scm-provider-gitexe>
+        <version.maven-gpg-plugin>1.6</version.maven-gpg-plugin>
+
+        <version.graphql-java-codegen>1.5.0-SNAPSHOT</version.graphql-java-codegen>
```
joffrey-bion, line 72: Same here, shouldn't this use the current version variable instead of a hardcoded version that happens to be equal to the current one?

plugins/graphql-java-codegen-gradle-plugin/graphql-codegen-gradle-plugin/build.gradle
```diff
@@ -0,0 +1,48 @@
+plugins {
```
joffrey-bion, line 1: Should this file be one level up? There is an empty `build.gradle` one level up. Is it normal?

CONTRIBUTING.md
kobylynskyi: That's correct. the module is named that way.

plugins/graphql-java-codegen-gradle-plugin/graphql-codegen-gradle-plugin/build.gradle
kobylynskyi: Thanks. That's a good point.
I've removed the hardcoded version wherever possible. In other places, I don't see how we can refer to an external file.

plugins/graphql-java-codegen-maven-plugin/pom.xml
kobylynskyi: Fixed.

plugins/graphql-java-codegen-gradle-plugin/graphql-codegen-gradle-plugin/build.gradle
kobylynskyi: I think it is totally normal. This set of plugins are required only by `graphql-codegen-gradle-plugin` module.

joffrey-bion:
Thanks for all the work you put in this!

PR #40, opened 2020-04-03
Add maven plugin parameters for directory support

joffrey-bion:
Currently working on the Gradle plugin as well, but it may be a bit longer.

Resolves:
https://github.com/kobylynskyi/graphql-java-codegen/issues/30


kobylynskyi:
@joffrey-bion awesome! ✌️
Would you be able to make the changes in Gradle plugin as well?

Issue #43, opened 2020-04-03
Improve project structure


kobylynskyi:
**Goal:**
Improve project structure.

The original request from @joffrey-bion in #41:

> The example project can't be built because it is commented out in settings.gradle. However, the plugin's project can't be built the first time if the other one is uncommented, because the dependency is not declared in the build.
Improve the projects setup so that the global build tests everything (including example projects) without having to follow a specific order and without changing directories.

PR #52, opened 2020-04-05
Improve project structure

### LINKED ISSUE #51: Configs in "graphqlSchemas" should be optional
If only `graphqlSchemaPaths` is specified then the `build` gradle task fails.

**Gradle config:**
```groovy
graphqlCodegen {
    graphqlSchemaPaths = ["$projectDir/src/main/resources/schema.graphqls".toString()]
    outputDir = new File("$buildDir/generated")
}
```

**Output:**
```
Bogdans-MacBook-Pro:gradle bogdankobylynskyi$ ./gradlew clean build publishToMavenLocal
> Task :example:graphqlCodegen FAILED

FAILURE: Build failed with an exception.

* What went wrong:
A problem was found with the configuration of task ':example:graphqlCodegen'.
> No value has been specified for property 'graphqlSchemas.rootDir'.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 1s
3 actionable tasks: 3 executed
```

kobylynskyi:
Resolves: #51 #43

After deep research on testing Gradle plugins, I was not able to find a solution where we can have a plugin and example modules in the same multi-module project. Here's some explanation from Gradle guys:
https://discuss.gradle.org/t/how-to-substitute-buildscript-dependency/23470/2

So in this PR, I have:
* Removed gradle wrapper files all over the place.
* Made plugin projects and example projects consistent across Maven/Gradle.
* Fixed #51


joffrey-bion:
Thanks for all the effort and for doing the research!

To be honest, I don't really see the point of nesting this much.
I'm more used to flat sub-project structures, with no project at the root, like:
- `graphql-java-codegen` (as a nested sub-project, not at the root)
- `graphql-java-codegen-gradle-plugin`
- `graphql-java-codegen-maven-plugin`
- `examples` (the 2 example projects can be nested there, since they're more for documentation purposes)
- a root `settings.gradle` listing at least the core lib and the gradle plugin (so that they're compiled together and the gradle plugin can have a local dependency on the core lib). In the future, if we manage to setup a gradle build for the maven plugin, that would allow to have a local dependency from the maven project to the core lib, which would be great as well.

Anyway, these are just my thoughts, and if you've got a working setup for now, this is absolutely great! I'm approving.

kobylynskyi:
Thanks for your feedback!

PR #54, opened 2020-04-06
Generate Builder for model classes #10

### LINKED ISSUE #10: Generate Builder for model classes
https://www.vogella.com/tutorials/DesignPatternBuilder/article.html

kobylynskyi:
Resolves #10

### DIFF
--- a/src/main/resources/templates/javaClassGraphqlType.ftl
+++ b/src/main/resources/templates/javaClassGraphqlType.ftl
@@ -58,6 +58,7 @@ public class ${className} <#if implements?has_content>implements <#list implemen
 </#if>
     }
 </#if>
+
 <#if toString>
     @Override
     public String toString() {
@@ -74,6 +75,31 @@ public class ${className} <#if implements?has_content>implements <#list implemen
 </#list>
 </#if>
         return joiner.toString();
+    }
+</#if>
+
+<#if builder>
+    public static class Builder {
+
+<#list fields as field>
+        private ${field.type} ${field.name}<#if field.defaultValue?has_content> = ${field.defaultValue}</#if>;
+</#list>
+
+        public Builder() {
+        }
+
+<#list fields as field>
+        public Builder set${field.name?cap_first}(${field.type} ${field.name}) {
+            this.${field.name} = ${field.name};
+            return this;
+        }
+
+</#list>
+
+        public ${className} build() {
+            return new ${className}(<#list fields as field>${field.name}<#if field_has_next>, </#if></#list>);
+        }
+
     }
 </#if>
 }
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,8 +18,8 @@ jobs:
           keys:
             - deps-{{checksum ".circle-cache-key"}}
       - run:
-          name: Build and run unit tests
-          command: gradle install
+          name: Build and publish to local Maven repo
+          command: gradle build publishToMavenLocal
       - save_cache:
           paths:
             - ~/.gradle
--- a/build.gradle
+++ b/build.gradle
@@ -122,5 +122,11 @@ publishing {
 }
 
 signing {
+    // this workaround should be obsolete in gradle 6.2: https://github.com/gradle/gradle/issues/11387
+    def version = String.valueOf(project.version)
+    if (version.endsWith("-SNAPSHOT")) {
+        project.version = version.substring(0, version.lastIndexOf("-SNAPSHOT"))
+    }
     sign publishing.publications.mavenJava
+    project.version = version
 }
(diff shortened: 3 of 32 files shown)


src/main/resources/templates/javaClassGraphqlType.ftl
```diff
@@ -76,4 +77,29 @@ public class ${className} <#if implements?has_content>implements <#list implemen
...
+<#list fields as field>
+        public Builder set${field.name?cap_first}(${field.type} ${field.name}) {
+            this.${field.name} = ${field.name};
+            return this;
+        }
+
+</#list>
+
+        public ${className} build() {
+            return new ${className}(<#list fields as field>${field.name}<#if field_has_next>, </#if></#list>);
```
valinha, line 100: Using a builder pattern would make sense that the constructor of the model class would be private, what do you think?

joffrey-bion: Not necessarily private, but it could take all fields as arguments, and fields could be final.

valinha: if you use a builder pattern it is not good practice to create instances with the constructor of the model class. Normally to avoid this, the model class will have a constructor with reduced visibility (package or private). Example:

public class Sample {
    private final String test;
    private final Integer test2;
    private final List<String> test3;

    Sample(final String test, final Integer test2, final List<String> test3) {
        this.test = test;
        this.test2 = test2;
        this.test3 = test3;
    }

    public static Sample.SampleBuilder builder() {
        return new Sample.SampleBuilder();
    }

    public static class SampleBuilder {
        private String test;

        SampleBuilder() {
        }

        public Sample.SampleBuilder test(final String test) {
            this.test = test;
            return this;
        }

        public Sample build() {
            return new Sample(this.test);
        }

        public String toString() {
            return "Sample.SampleBuilder(test=" + this.test + ")";
        }
    }
}

This is the default approach of Lombok @Builder for example.

If you put all fields final you have to delete the default constructor.

Maybe it's good to parameterize both options? @joffrey-bion.

Note: If we put the constructor private or package visibility maybe the default value of generateBuilder would be false for compatibilty.

ReneBL: I think the approach @valinha suggests is the most appropiated in this case. If you use a builder pattern, it's mainly because you want to take control over the instantiation of an object. That implies reducing the visibility of the constructor to package or private.

@joffrey-bion for me it's ok to take all fields as arguments and have them final, but you also need what i'm saying in the previous paragraph.

kobylynskyi: @valinha @joffrey-bion @ReneBL, thank you guys for your review!
So I think we have 2 options here:
1. Make `generateBuilder=false` by default. And if it is `true` then reduce the visibility of constructors to package-level and make all fields as final.
2. Make `generateBuilder=true` by default. And in order to be backward-compatible, we will need to keep no-args and all-args constructors as public. Fields then will be not final.

Keep in mind that we have a single global configuration `generateBuilder` for **all** model classes. So imagine a case when there's a class A with 30 fields and a class B with 1 field. For me, it will be easier to use a builder for class A and all-args-constructor for class B because of less code. (instead of `new B.Builder().setField(field).build()` it will be just `new B(field)`).

What do you guys think?

joffrey-bion: >  If you use a builder pattern, it's mainly because you want to take control over the instantiation of an object. That implies reducing the visibility of the constructor to package or private.

I don't think this is the main usage of the builder pattern. If you want to control instantiations, as in particular constructor calls or even particular implementations, then a factory is more appropriate.

I believe the builder pattern is useful when a class enforces some invariants on all instances (like immutability or constraints involving dependent fields), and does this in its constructor. In this setup, it might not be easy to satisfy all constraints in one go when creating a new instance using its constructor, and we might want to pass an incomplete object around a little before finalizing its instantiation.
To me that's the main goal of the builder, and it doesn't imply hiding constructors at all, because it's perfectly possible to build the object in one go in some situations, and there is no need to force 2 instantiations upon the user of the class in all situations.

joffrey-bion: Coming back to what @kobylynskyi said, I would vote for option 2: keeping the visibility public for constructors, and only generate builders on top of that with the `generateBuilder` option.

We could then add more things later, like `generateImmutableClasses` (which would imply generating builders).

valinha: I think it's good to cover this in the future with the generateImmutableClasses option (false by default) that generates the builder and removes the empty constructor.
Another solutions like GRPC create immutable classes, example:
https://developers.google.com/protocol-buffers/docs/reference/java-generated

Issue #37, opened 2020-04-03
Client-side code generation


kobylynskyi:
**Is your feature request related to a problem? Please describe.**
Currently, the main purpose of the plugin is to generate server-side code (POJOs and interfaces). 
The idea is to add support of client-side classes generation. This will allow users to build a GraphQL request.

**Describe the solution you'd like**
Schema example:
```graphql
type Mutation {
    newBike(bikeInput: BikeInput!): Bike
}

input BikeInput {
    type: BikeType!
    brand: String!
    size: String!
    price: BigDecimal
}

type Bike {
    id: ID
    type: BikeType
    brand: String
    size: String
    price: BigDecimal
    addedDateTime: DateTime
}
```

Generated class:

```java
public class NewBikeMutationRequest {
    private BikeInput bikeInput;
    //getters/setters
}

public class BikeInputTO {
    private BikeTypeTO type;
    private String brand;
    private String size;
    private String price;

    public BikeInputTO() { }

    //getters/setters
```

The end-user will then use these classes to build the request and **response projection** (open question for now). So that it will be serialized to the following request string:

```graphql
mutation newBike {
  newBike(bike: {
    type: ROAD,
    brand: "LOL",
    size: "Big",
    year: 2019,
    price: "123.45"
  }) {
    id
    addedDateTime
  }
}
```

kobylynskyi:
@joffrey-bion 
I was not able to find a solution for client-side code generation in Java.

The idea of this feature would be:
1. Define the projection of the response object (which fields to fetch)
2. Properly serialize the request.

So the code on the client-side will look like:
```java
BikeInputTO input = new BikeInputTO(...);
NewBikeMutationRequest request = new NewBikeMutationRequest(input)
                .responseProjection(new BikeProjection()
                        .id().addedDateTime());
Bike bikeResponse = restTemplate.postForObject("https://localhost:8080", request, Bike.class);
```
(the DSL is still discussable)

joffrey-bion:
> I was not able to find a solution for client-side code generation in Java.

Yes, I was actually mentioning code generators in general.
The one I heard about was for Swift I think.

> The idea of this feature would be:
> 1. Define the projection of the response object (which fields to fetch)
> 2. Properly serialize the request.

Oh OK, so you want to build client queries in Java directly?
That could be a way too, I just felt like the idea of this library was to go schema-first, so I expected a query-first solution here. But that implies parsing queries. I don't know if the library you're using can do this.

[...]

PR #59, opened 2020-04-08
Fix GraphQL query serialization and add Gradle plugin for client-codegen #37

kobylynskyi:
1. Add client-code configs to the Gradle plugin.
2. Make 2 separate example projects: `example-client` and `example-server`
3. Fixed characters escaping as suggested by @joffrey-bion


src/main/java/com/kobylynskyi/graphql/codegen/model/request/GraphQLRequestSerializer.java
```diff
@@ -37,8 +36,8 @@ public static String serialize(GraphQLRequest graphQLRequest) {
             builder.append(graphQLRequest.getResponseProjection().toString());
         }
         builder.append(" }");
-        builder.append("\"}");
-        return builder.toString();
+        String escapedString = builder.toString().replaceAll("\"", "\\\\\"");
```
joffrey-bion, line 39: This is probably insufficient, what about all other string escapes? If the query contains input strings with unicode character escapes etc. I guess a starting point would be to escape not only quotes but also backslashes.

Second point, why the triple escape? We're escaping things that are part of the raw query (like `"`) so that it can be inserted into the JSON, so `\"`. That is one level. Now to write this escape as a Java literal, so we need to escape this backslash and quote, so `\\\"`. Why the third level?

kobylynskyi: Could you please give an example of such query?
Because based on the queries that I tested - everything worked as expected

kobylynskyi: Could you please give an example of such query?
Because based on the queries that I tested - everything worked as expected

joffrey-bion: Sure, let's assume the following schema:
```
type Query {
    autocomplete(inputStr: String!): [String]!
}
```
We can write this query:
```
query AutoComplete {
    autocomplete(inputStr: "\u1234\"\n")
}
```
This means that the actual JSON to send is the following:
```
{
    "query": "query AutoComplete {\n    autocomplete(inputStr: \"\\u1234\\\"\\n\")\n}"
}
```

kobylynskyi: Ok. Thank you so much. 
Will test and let you know.

kobylynskyi: @joffrey-bion you were right. special characters and such things like `\n`, '\t' were a problem.
And looks like `StringEscapeUtils.escapeJava();` does the trick.

kobylynskyi:
@joffrey-bion could you please cross-check this PR? Thank you!

src/test/resources/expected-classes/request/graphql-query/eventsByIdsQuery.txt
```diff
@@ -1,5 +1,5 @@
 query {
-  eventsByIds(ids: [ "4", "5", "6" ]){
+  eventsByIds(ids: [ \"\u1234\", \"\"\", \"\n\" ]){
```
joffrey-bion, line 2: I believe here the query should not escape the outermost quotes of each string, otherwise it's probably invalid from a GraphQL standpoint. We expect `ids` to be of type `[String]` here, so we should really provide a list of double-quoted strings.

src/test/resources/expected-classes/request/graphql-query/updateIssueMutation.txt
```diff
@@ -3,7 +3,7 @@ mutation {
     floatVal: 1.23,
     booleanVal: false,
     intVal: 42,
-    stringVal: "my-default",
+    stringVal: \"my-default\",
```
joffrey-bion, line 6: Same here, the query itself expects these quotes, so I believe they should not be escaped.

src/main/java/com/kobylynskyi/graphql/codegen/model/graphql/GraphQLRequestSerializer.java
```diff
@@ -37,8 +38,7 @@ public static String serialize(GraphQLRequest graphQLRequest) {
             builder.append(graphQLRequest.getResponseProjection().toString());
         }
         builder.append(" }");
-        builder.append("\"}");
-        return builder.toString();
+        return String.format("{\"query\":\"%s\"}", StringEscapeUtils.escapeJava(builder.toString()));
```
joffrey-bion, line 41: What we technically want is to escape strings for JSON, not for Java, so we should respect the JSON specification instead, which is not the contract of `StringEscapeUtils.escapeJava()`.
That's why I was mentioning that maybe we should use a JSON library for this purpose (or ask the user to provide one).
As a temporary workaround though, this can work.

src/main/java/com/kobylynskyi/graphql/codegen/model/graphql/GraphQLRequestSerializer.java
```diff
@@ -37,8 +38,7 @@ public static String serialize(GraphQLRequest graphQLRequest) {
             builder.append(graphQLRequest.getResponseProjection().toString());
```
joffrey-bion, line 38: Do we expect this `toString()` here to properly encode string literals respecting the GraphQL specification?
http://spec.graphql.org/June2018/#sec-Appendix-Grammar-Summary.Lexical-Tokens

I think this requires quite some work to be done properly. We should probably build an actual query data model (or use a potentially internal one from `graphql-java`) defining all the elements that constitute a GraphQL query/mutation like a query field, nested field, input parameters, directives etc.

joffrey-bion:
I'm sorry for all these comments, I hope it doesn't sound like a rant or anything.
I'm really just trying to see what could go wrong here.
I believe implementing client code generation is a complicated task, and I wonder if we want to have something quickly working for most use cases, or if we want to really spend the time to cover all use cases and match the specifications.

src/main/java/com/kobylynskyi/graphql/codegen/model/graphql/GraphQLRequestSerializer.java
kobylynskyi: @joffrey-bion Thank you so much for your input.
I've utilized Jackson library because we are already using it to parse external mapping config from son file (see `JsonMappingConfigSupplier`).

src/test/resources/expected-classes/request/graphql-query/updateIssueMutation.txt
kobylynskyi: @joffrey-bion Fixed

src/test/resources/expected-classes/request/graphql-query/eventsByIdsQuery.txt
kobylynskyi: @joffrey-bion Fixed. Thank you!

src/main/java/com/kobylynskyi/graphql/codegen/model/graphql/GraphQLRequestSerializer.java
kobylynskyi: @joffrey-bion That's a very valid point. I didn't pay attention to it.
I've refactored the whole serialization logic. Could you please take a look again?

kobylynskyi:
@joffrey-bion I appreciated your effort in reviewing this.
All the comments as of now were addressed.
Here's what the testing looked like:
1. GraphQL client querying special characters:
Screenshot 2020-04-09 11 47 03

2. GraphQL server receiving them:
Screenshot 2020-04-09 11 46 46

PR #62, opened 2020-04-09
Maven plugin support for client code generation + example projects #37

kobylynskyi:


plugins/maven/graphql-java-codegen-maven-plugin/src/main/java/io/github/kobylynskyi/graphql/codegen/GraphqlCodegenMojo.java
```diff
@@ -79,6 +79,15 @@
...
     private Set<String> fieldsWithResolvers = new HashSet<>();
 
+    @Parameter(defaultValue = "false")
+    private boolean generateRequests;
+
+    @Parameter(defaultValue = "Request")
+    private String requestSuffix;
+
+    @Parameter(defaultValue = "ResponseProjection")
+    private String responseProjectionSuffix;
```
joffrey-bion, line 89: Maybe the client-related config could be declared in a separate config (nested object clientGeneration or something like this). Or at least we should really make it clear in the doc.
If we do this, then maybe the server-related config like the API/async stuff should also be in its own config.
The common part could stay top level.

kobylynskyi: Yeah, I also was thinking about it. Maybe will do it in the next PR.
Regarding documentation - I had an idea to make the separate markdown file with a detailed explanation and examples of each config. Just to make README more clear.

Issue #64, opened 2020-04-10
Unify operations and field resolvers data and templates


joffrey-bion:
Conceptually, GraphQL operations (queries/mutations/subscriptions) are a special case of field resolvers that happen to be on well known root types.

Technically, this means that everything about field resolvers is the same as these root types resolvers, apart from the fact that the former get an instance of the resolved type as first param.

Therefore, all the configuration that applies to field resolvers also applies to resolvers of the fields that are part of these root types, like `CompletableFuture` return type, or `DataFetcherEnvironment` parameter.

This is why this issue is about unifying the data model and the templates of operations and field resolvers. This will allow to solve both the following issues at the same time:
https://github.com/kobylynskyi/graphql-java-codegen/issues/61
https://github.com/kobylynskyi/graphql-java-codegen/issues/63

Issue #63, opened 2020-04-10
Field resolvers should also be subject to async APIs configuration


joffrey-bion:
## Issue Description
The generated field resolvers currently don't respect the `generateAsyncApis` configuration parameter, although they are part of the API just like top-level resolvers.

## Expected Result
When `generateAsyncApis` is enabled, field resolvers should have a return type wrapped in `CompletableFuture` as well.

## Actual Result
Field resolvers just have a plain java type as return type.

vinaybedre:
@kobylynskyi I dont see any documentation on how to use `generateAsyncApis`
Can you please let me know if this option is supported?

joffrey-bion:
@vinaybedre I think for more flexibility, this configuration has since been changed to `apiReturnType`, so you can specify any kind of return type, such as `CompletableFuture`:

This is what I have in my current Maven config:
```
<apiReturnType>java.util.concurrent.CompletableFuture</apiReturnType>
```

vinaybedre:
@joffrey-bion Thanks for the hint.
Do you know is it possible to change return type of a specific field resolver?

joffrey-bion:
@vinaybedre I don't believe this is possible at the moment. But I'll let @kobylynskyi confirm that, as I haven't contributed here in a while

Issue #61, opened 2020-04-08
Add DataFetchingEnvironment parameter to main query resolver


joffrey-bion:
**Is your feature request related to a problem? Please describe.**
When using GraphQL Java Tools, the main Query resolver, like other resolvers, can take an optional `DataFetchingEnvironment` parameter.
It is currently generated in all resolvers interfaces, but not for the root resolvers like `Query`.

**Describe the solution you'd like**
It'd be nice to provide this `DataFetchingEnvironment` argument to the query resolver's methods.
If we need backwards compatibility, we could use a config parameter to add or not the DataFetchingEnvironment argument in the generated interfaces.

**Describe alternatives you've considered**
There is no other way currently than to stop implementing the `Query` interface, which is a pain because changes to the schema don't make the compilation fail.

kobylynskyi:
Good idea.
I think we definitely need a configuration parameter for this (which will default to ‘false’)

joffrey-bion:
Should this param also affect field resolvers?

kobylynskyi:
Yes, agree.

joffrey-bion:
I tried to look into this one, but I realized that there is some duplication around operations VS field resolvers. They are currently treated differently, but the root Query object is just a special case of type that only needs a resolver and no POJO.

Technically `OperationDefinition` could be the same thing as `FieldResolverDefinition` (and thus would automatically solve the current issue). Would you like me to make this unification? Or do you prefer to keep changes to a minimum and maybe unify later as a separate effort?

kobylynskyi:
Yes, we could definitely make the unification. Thanks for finding this.
But how will this automatically solve the issue?

joffrey-bion:
By automatically I meant that if the code is not duplicated, adding the
DataFetchingEnvironment in one place solves the other one.

kobylynskyi:
@joffrey-bion, will you be able to work on this feature? I would really appreciate it.
Otherwise, I can do it over the weekend.

joffrey-bion:
Yes, I'll try to work on it tomorrow

PR #66, opened 2020-04-10
Unify data and template of field resolvers and operations

joffrey-bion:
Conceptually, GraphQL operations (queries/mutations/subscriptions) are a
special case of field resolvers that happen to be on well known root types.

Technically, this means that everything about field resolvers is the same
as these root types resolvers, apart from the fact that the former get an
instance of the resolved type as first param.

Therefore, all the configuration that applies to field resolvers also
applies to resolvers of the fields that are part of these root types, like
CompletableFuture return type, or DataFetcherEnvironment parameter.

Resolves: https://github.com/kobylynskyi/graphql-java-codegen/issues/64
Resolves: https://github.com/kobylynskyi/graphql-java-codegen/issues/61
Resolves: https://github.com/kobylynskyi/graphql-java-codegen/issues/63


joffrey-bion:
@kobylynskyi thanks for the feedback, yes the plugin changes were on the way :)

src/test/resources/expected-classes/CommitResolver.java.txt
```diff
@@ -7,14 +7,19 @@ public interface CommitResolver {
 
     PullRequestConnection associatedPullRequests(Commit commit, String after, String before, Integer first, Integer last, PullRequestOrder orderBy, DataFetchingEnvironment env) throws Exception;
 
+    @javax.validation.constraints.NotNull
+    @com.fasterxml.jackson.databind.annotation.JsonDeserialize(using = com.example.json.DateTimeScalarDeserializer.class)
```
kobylynskyi, line 11: @joffrey-bion I think its a bit late to ask but why do we have this annotation here?

kobylynskyi:
@joffrey-bion I think we can release 1.6.0 and this will be the last changes
Please let me know and I will release tmrw morning (US time)

joffrey-bion:
That's great news! About the annotation, it's just matching the config that was done in the test. It also bugged me at first 😄 I guess it was really just a test to check if the annotation was put correctly, not an actual use case, because I see no reason for Blame to be a date

Issue #81, opened 2020-04-20
Add support for extensions


joffrey-bion:
**Is your feature request related to a problem? Please describe.**

[Type extensions](http://spec.graphql.org/draft/#TypeExtension) are part of the specification, and should therefore be supported. Type extensions, according to the linked spec, can be of the following types:

- ScalarTypeExtension (`extend scalar X`)
- [ObjectTypeExtension](http://spec.graphql.org/draft/#ObjectTypeExtension) (`extend type X`)
- InterfaceTypeExtension (`extend interface X`)
- UnionTypeExtension (`extend union X`)
- EnumTypeExtension (`extend enum X`)
- InputObjectTypeExtension (`extend input X`)

**Describe the solution you'd like**

When defining type extensions in the SDL schema, they should be taken into account and "merged" with the extended type so that the code generation matches the actual schema.

We could (should?) also provide an option (`generateExtensionFieldsResolvers`?) to consider all fields from Object Type Extensions as fields with resolvers, because it is likely that these fields require a special resolver implementation with a service call. (Should it be true by default? Maybe not.)

If we do this, maybe we should add a new option to do the opposite of `fieldsWithResolvers`, like `fieldsWithoutResolvers`, blacklisting some fields so that they are part of the generated POJO instead of being generated as part of the resolver interface.
The reason why I think we should do this is because if the user decides to generate resolvers for *most* fields that are part of extensions, but not all, it would be painful to list all of them in `fieldsWithResolvers`. Instead, we should allow declaring `generateExtensionFieldsResolvers` to handle the general case, and add some exceptions via `fieldsWithoutResolvers`.

**Describe alternatives you've considered**

The alternative (current situation) is to force the user to declare all their types in the same schema, which is not ideal and sometimes impossible (in case of remote schema using federation).

Lukas-Kullmann:
I ran into the same issue today when I wanted to extend an enum type. So when implementing this, it would be super nice if all type extensions would be supported.

- `extend scalar X`
- `extend type X`
- `extend interface X`
- `extend union X`
- `extend enum X`
- `extend input X`

I just wanted to mention this since the issue description suggests that there is only `extend type X`.

kobylynskyi:
Thanks for your feature request. 
Current implementation of the library is not supporting extensions.
I will work on this soon.

joffrey-bion:
[...]

PR #82, opened 2020-04-21
Support extensions for all GraphQL types #81

kobylynskyi:
It is the first part of the changes related to extensions.

As suggested by @joffrey-bion, upcoming changes will include:
* Config `generateExtensionFieldsResolvers`
* Config `fieldsWithoutResolvers`

### DIFF
--- a/src/main/java/com/kobylynskyi/graphql/codegen/mapper/MapperUtils.java
+++ b/src/main/java/com/kobylynskyi/graphql/codegen/mapper/MapperUtils.java
@@ -3,6 +3,7 @@ package com.kobylynskyi.graphql.codegen.mapper;
 import com.kobylynskyi.graphql.codegen.model.MappingConfig;
 import com.kobylynskyi.graphql.codegen.model.graphql.GraphQLOperation;
 import com.kobylynskyi.graphql.codegen.utils.Utils;
+import graphql.language.Definition;
 import graphql.language.Document;
 import graphql.language.NamedNode;
 import graphql.language.UnionTypeDefinition;
@@ -13,7 +14,7 @@ import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
 
-public class MapperUtils {
+class MapperUtils {
 
     private static final Set<String> JAVA_RESTRICTED_KEYWORDS = new HashSet<>(Arrays.asList(
             "abstract", "assert", "boolean", "break", "byte", "case", "catch", "char", "class", "const", "continue",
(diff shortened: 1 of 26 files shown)


src/main/java/com/kobylynskyi/graphql/codegen/model/DefinitionTypeDeterminer.java
```diff
@@ -6,27 +6,35 @@
...
+        } else if (definitionClass.equals(UnionTypeDefinition.class)) {
             return GraphqlDefinitionType.UNION;
-        } else if (definition instanceof ScalarTypeDefinition) {
+        } else if (definitionClass.equals(ScalarTypeDefinition.class)) {
             return GraphqlDefinitionType.SCALAR;
-        } else if (definition instanceof InterfaceTypeDefinition) {
+        } else if (definitionClass.equals(InterfaceTypeDefinition.class)) {
             return GraphqlDefinitionType.INTERFACE;
-        } else if (definition instanceof DirectiveDefinition) {
+        } else if (definitionClass.equals(DirectiveDefinition.class)) {
```
joffrey-bion, line 37: Rather than this big `if-else` sequence with class checks, why not use directly the `Document.getDeclarationsOfType(...)` to get all declarations of a given class?
We wouldn't have to *then* check which kind of declaration it is, we would just ask for the ones we want at each step.

kobylynskyi: It will slow down the generation process.
Because if you look at ‘getDeclarationsOfType’ method, they iterate through all definitions.
So I think our approach is a liiittle better from performance point of view.

joffrey-bion: Fair enough. For huge schemas this might have some significant overhead.

src/main/java/com/kobylynskyi/graphql/codegen/mapper/MapperUtils.java
```diff
@@ -46,30 +47,51 @@ static String capitalizeIfRestricted(String fieldName) {
...
+     *
+     * @param definitionClass the definition class
+     * @param definitionName  the definition name
+     * @param <T>             the type of definition
+     * @return a list of definitions of that class/name or empty list
+     */
+    static <T extends Definition<?>> List<T> getDefinitionsOfType(Document document,
+                                                                  Class<T> definitionClass,
+                                                                  String definitionName) {
+        return document.getDefinitions().stream()
```
joffrey-bion, line 75: Wait, I thought we were saving on `getDefinitions()` iterations for the sake of performance, but now we're iterating many times over it here.

I believe we should instead process the whole document once and build an intermediate representation that suits our needs, and then use this intermediate representation to generate classes/interfaces.
This way it would clearly separate the code processing the schema (which may need to change when `graphql-jaav` is updated) and the code building the data model for the generation.

joffrey-bion: We could even define only one such class with generic types:

```java
class ExtendedDefinition<B, E> {
    private B baseDefinition;
    private Collection<E> extensions;

    // getters/setters/constructors
}
```

kobylynskyi: Thanks for your feedback. You are right.
But in order to make this PR clear, let me do this performance improvement in the next PR.
Thank you!

joffrey-bion: Definitely.

PR #87, opened 2020-04-22
Refactor document parsing logic

### LINKED ISSUE #86: Performance improvement: scan the document once
Currently we are iterating through the whole document multiple times in order to retrieve extensions, implemented interfaces, unions having the particular type, etc.

Scan the GraphQL document for definitions once in `GraphQLDocumentParser`.

kobylynskyi:
Resolves #86


src/main/java/com/kobylynskyi/graphql/codegen/model/definitions/ExtendedEnumTypeDefinition.java
```diff
@@ -0,0 +1,23 @@
...
+    /**
+     * Get enum value definitions from the definition and its extensions
+     *
+     * @return list of all enum value definitions
+     */
+    public List<EnumValueDefinition> getValueDefinitions() {
+        List<EnumValueDefinition> definitions = definition.getEnumValueDefinitions();
+        extensions.stream()
+                .map(EnumTypeExtensionDefinition::getEnumValueDefinitions)
+                .forEach(definitions::addAll);
```
joffrey-bion, line 20: I'm not sure how safe this is. Is `definition.getEnumValueDefinitions()` creating a defensive copy of the list internally?
To be on the safe side, I'd rather create a list on the spot add stuff from the base definition and extensions.

kobylynskyi: Yes, they are creating a defense copy.

Issue #83, opened 2020-04-21
codegen doesn't build consistently


jyjyj:
## Issue Description
I'm using Eclipse and gradle. Whenever I update my graphql schema, I expect to either:

Option 1
1) Refresh Gradle Project
2) Clean project
3) then Refresh project

or

Option 2
1) gradle build

to refresh the generated code (classes and interfaces). My issues are that I haven't been able to find a consistent method of re-generating the Java code after updating the schema. Sometimes I do step (1) and it works and other times it doesn't, same for trying step (2), and at the moment neither of these work. 

## Steps to Reproduce
Can't consistently do it as explained above.

## Expected Result
I'm expecting either step (1) or (2) to correctly re-generate the classes, and hence am flagging it as a bug.

## Actual Result
What I have found as a workaround is to:

1) delete the folder with the generated code
2) run gradle build again
3) refresh gradle project

and the new code comes up. This seems a little strange, and I'm not expecting to need to delete and rebuild every single time I make some changes in my schema.

If I'm wrong and there's actually a proper way to re-generate the Java code consistently, or if delete + rebuild is actually intended, please point me in the right direction! Thanks :)

joffrey-bion:
Thanks for opening this as a separate issue.
As of now, it lacks crucial information and I'm not sure @kobylynskyi or I can be of much help here.

1. Please define "does not work":
   - Is it a Gradle build failure?
   - Is it an IDE error/crash when rebuilding the project from it?
   - Are the expected generated classes actually missing?
   - Are the expected generated classes present but not seen by the IDE?
  These are very different issues, that may or may not be addressed by a fix in this library.

2. Please provide the exact error you get from the build (if it's a Gradle failure), or explain/give a screenshot of your IDE error

3. Please provide an example schema and Gradle build file. Even if we can't reproduce consistently, maybe we can help spotting potential problems.

Without this kind of information, we are absolutely clueless as we don't even know what doesn't work on your machine.

jyjyj:
Hi,

Now that I have separated this from issue #81, I do not see any errors when building gradle, nor does the IDE crash or have any errors. As for whether the classes are missing or present but not seen by the IDE, I have monitored the folder where the code is being generated into, and it does not update when I make a change to the schema.

[...]

PR #91, opened 2020-04-23
Gradle plugin: fix classes regeneration when schema changes #83

kobylynskyi:
Gradle cannot identify whether incoming graphql schema files were changed if using SchemaFinder.
So setting `outputs.upToDateWhen` to `false` in order to always regenerate java classes.


joffrey-bion:
This seems pretty overkill IMO. I approve the PR so that we have a quick workaround, but I believe we can actually fix it.

I believe we can run the schema finder every time and then check the resulting list of paths as an `@InputFiles` (which check contents).

This can be done by defining a plugin extension (which is [the classic way](https://docs.gradle.org/current/userguide/custom_plugins.html#sec:getting_input_from_the_build) of configuring a plugin) instead of directly writing to task inputs, but maybe we can even just declare the getter for schema paths public and annotate it `@InputFiles`.

I'm trying to see if the second one works or if we would need the extension.

kobylynskyi:
@joffrey-bion this was indeed a quick dorty hack.
Thank you for actually fixing this in #92

Issue #84, opened 2020-04-21
Generate JavaDoc matching the schema comments


joffrey-bion:
Currently, no documentation is generated in the classes even when the schema contains doc comments.
I'm not sure the graphQL parser provides access to comments, but if it does, it'd be great to make them appear as Javadoc.

kobylynskyi:
Will be released in version 1.7.0 this weekend.

PR #99, opened 2020-04-24
Add javadoc to all generated classes #84

kobylynskyi:


src/main/resources/templates/javaClassGraphqlType.ftl
```diff
@@ -27,9 +41,19 @@ public class ${className} <#if implements?has_content>implements <#list implemen
 </#if>
 
 <#list fields as field>
+<#if field.javaDoc?has_content>
+    /**
+     * {@link ${className}#${field.name}}
```
joffrey-bion, line 46: I'm not sure about this. From a user's perspective, when we use the generated code, we mostly use getters and setters (we can't access the private fields), so this will add one level of indirection that may not be desirable.
I think we could simply use the doc of the field as-is on the getter and on the setter.
What do you think?

kobylynskyi: I just didn't want to duplicate the javadoc in two places

joffrey-bion: It doesn't really matter much in generated code I think.
Even in normal code, I would want to write a decent Javadoc on the getter and on the setter, or at the very least on the getter, because private fields are not accessible and therefore their doc.

kobylynskyi: I think then maybe it is better then to have on the getter and do the reference in the setter:
`{@link ${className}#set${field.name?cap_first}}`

joffrey-bion: I still don't see the benefit of this, compared to duplicate doc. When people will use the setter, we can either show them the doc directly or force them to click once more to go to the getter.

kobylynskyi: Done as per your suggestion

Issue #97, opened 2020-04-24
Preferred extension for schemas is .graphql


benmccann:
It looks to me like `.graphql` is the most commonly used extension for schemas: https://github.com/graphql/graphql-spec/issues/203#issuecomment-242140024

Could the project support either just `.graphql` or both `.graphql` and `.graphqls` by default?

kobylynskyi:
@benmccann 
the codegen library works with both `.graphql` and `.graphqls`. 
You can just specify it in the maven config:
```xml
<graphqlSchemaPaths>
    <graphqlSchemaPath>${project.basedir}/src/main/resources/schema.graphql</graphqlSchemaPath>
</graphqlSchemaPaths>
```
or:
```xml
<graphqlSchemas>
    <includePattern>.*\.graphql</includePattern>
</graphqlSchemas>
```

benmccann:
Yeah, I was just suggesting a different default to better align with the ecosystem

kobylynskyi:
Oh, ok. You are right.
I will change the default value of `graphqlSchemas.includePattern` to support both `.graphql` and `.graphqls`

joffrey-bion:
It seems from what I have seen that people use .graphql for query files,
and .graphqls for schema files. That's also what my IDE picked up when
using these files.

But maybe I didn't push the research enough. I'd be happy to change the
default if it reflects better the convention, but I'm not sure it's
actually the case right now.

kobylynskyi:
The change will be released in v1.7.0 by the end of this week.

benmccann:
Thanks!

PR #100, opened 2020-04-24
Extend default SchemaFinder include pattern config #97

kobylynskyi:
Support both .graphqls and .graphql

### DIFF
--- a/src/main/java/com/kobylynskyi/graphql/codegen/supplier/SchemaFinder.java
+++ b/src/main/java/com/kobylynskyi/graphql/codegen/supplier/SchemaFinder.java
@@ -15,7 +15,7 @@ import java.util.stream.Stream;
  */
 public class SchemaFinder {
 
-    public static final String DEFAULT_INCLUDE_PATTERN = ".*\\.graphqls";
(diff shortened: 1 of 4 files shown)


src/main/java/com/kobylynskyi/graphql/codegen/supplier/SchemaFinder.java
```diff
@@ -15,7 +15,7 @@
  */
 public class SchemaFinder {
 
-    public static final String DEFAULT_INCLUDE_PATTERN = ".*\\.graphqls";
+    public static final String DEFAULT_INCLUDE_PATTERN = ".*\\.graphql[s]?";
```
joffrey-bion, line 18: This does work, but the `[]` in regex are normally used for character classes ("any of these characters").
Here we're defining a character class containing only one letter, so it's not really necessary, `.*\\.graphqls?` should be sufficient.
That being said, maybe we can use `()` for readability purposes.

kobylynskyi: Changed as per your suggestion.
Thank you, @joffrey-bion !

PR #109, opened 2020-04-25
Return generated files in GraphQLCodegen.generate #107

### LINKED ISSUE #107: GraphQLCodegen.generate should return generated List<File>
It looks like sbt requires source generators to return the list of files that the generated. I have most of a plugin done except for that

kobylynskyi:
@benmccann could you please review if that's what you need for the sbt plugin?

Resolves #107


benmccann:
Perfect! Thanks so much for the amazingly fast turnaround! :-)

PR #119, opened 2020-04-25
Local development workflow for sbt plugin

benmccann:


benmccann:
I might have just discovered a problem. Wait a minute before merging while I do some more testing

benmccann:
Ok, this is as good as we're going to do. This should really be setup slightly differently, but there's a bug in Gradle that makes it not feasible to do correctly. This will be fine to merge for now. I'll report an issue to the Gradle team and we can come back and clean this up later if it ever gets fixed

kobylynskyi:
Ok. Please share the issue when you created it.
Thanks!

PR #122, opened 2020-04-26
Add more options to the sbt plugin

benmccann:


benmccann:
That's what they suggest in the sbt docs: https://www.scala-sbt.org/1.x/docs/Plugins-Best-Practices.html#Key+naming+convention%3A+Use+prefix

PR #125, opened 2020-04-26
Setup build files for releasing plugin

benmccann:
The plugin should be ready to be released after https://github.com/kobylynskyi/graphql-java-codegen/pull/127 is merged and the account is setup as described in [the sbt docs](https://www.scala-sbt.org/1.x/docs/Bintray-For-Plugins.html)

I believe the steps for each release will simply be:
* Release core library (the sbt plugin depends on the core library with same version number)
* Run `sbt publish`

I'm not quite sure how to test that this is setup correctly. I guess we will find out the first time we try

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -52,7 +52,7 @@ Please follow the steps below in order to make the changes:
 
 ### For the sbt plugin
 
-1. Publish the core library locally with `./gradlew publishIvyPublicationToIvyRepository`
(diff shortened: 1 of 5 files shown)


benmccann:
I'm not sure why codecov didn't finish, but the GitHub code coverage finished. And I mostly updated sbt files, so it seems unrelated to this PR

kobylynskyi:
@benmccann Awesome! Thank you for the instructions!

PR #127, opened 2020-04-26
Add sbt plugin integration test

benmccann:
The test is failing, but I figured it's worth committing anyway to add the infrastructure. I'll probably need some help from other sbt devs to figure out what's wrong and having it checked in will help with that. I also fixed two issues that it found

One thing that it'd be good to get some review on - are the options that I changed to be of type `Option[String]` all optional? (i.e. are they all nullable in the Maven/Gradle plugins?)


benmccann:
The integration test is now passing and the plugin should be ready for prime time!

plugins/sbt/src/main/scala/io/github/kobylynskyi/graphql/codegen/GraphQLCodegenSbtPlugin.scala
```diff
@@ -88,51 +74,72 @@ object GraphQLCodegenSbtPlugin extends AutoPlugin {
...
+    inConfig(Compile)(baseGraphQLSettings)
+    // For now we skip compiling schemas under src/test because the plugin will
+    // fail if no schemas are found at the expected location
+    // ++ inConfig(Test)(baseGraphQLSettings)
 }
 
 object Codegen {
   def apply(
       outputDir: File,
       graphqlSchemaPaths: Seq[String],
```
kobylynskyi, line 107: The only required attribute should be `outputDir`. All other attributes should be optional.

benmccann: Thanks for taking a look!

`graphqlSchemaPaths` is optional already since you can pass an empty list, which I think would be the best way to handle that one

For `schemasRootDir`, I just changed the methods using it to take an `Option[File]` so that I remember that it's optional. There are still some options from the Maven and Gradle plugins that haven't been implemented. `schemasRootDir` is one of them. I haven't added a check for it yet because it's currently always provided until I get around to implementing support for the rest of the options

PR #129, opened 2020-04-28
Handle duplicate fields from the type and interface #128

### LINKED ISSUE #128: Code generation is broken with comments on interfaces' fields (Version: 1.7.0)
## Issue Description

In version 1.7, generating code for this example (based on [example No 62 from the spec](http://spec.graphql.org/June2018/#example-ab5e5)), will result in an error:

```graphql
interface NamedEntity {
  # an awesome name!
  name: String
}

type Person implements NamedEntity {
  name: String

  # age in years
  age: Int
}
```

## Expected Result
The code is generated correctly

## Actual Result
An error occurrs during compilation:

```
Error:(8,20) java: variable name is already defined in class com.example.Person
Error:(13,52) java: variable name is already defined in constructor Person(java.lang.String,java.lang.Integer,java.lang.String)
Error:(42,19) java: method getName() is already defined in class com.example.Person
Error:(48,17) java: method setName(java.lang.String) is already defined in class com.example.Person
Error:(58,24) java: variable name is already defined in class com.example.Person.Builder
Error:(79,24) java: method setName(java.lang.String) is already defined in class com.example.Person.Builder
```

The generated code is:

NamedEntity.java

```java
package com.example.api;


public interface NamedEntity {

    /**
     * an awesome name!
     */
    String getName();

}
```

Person.java
**Note** that `name` is generated twice because of the comment in the interface.
If I remove the comment in the interface, it works fine.

```java
package com.example.api;


public class Person implements NamedEntity{

    private String name;
    private Integer age;
    private String name;

    public Person() {
    }

    public Person(String name, Integer age, String name) {
        this.name = name;
        this.age = age;
        this.name = name;
    }

    public String getName() {
        return name;
    }
    public void setName(String name) {
        this.name = name;
    }

    /**
     * age in years
     */
    public Integer getAge() {
        return age;
    }
    /**
     * age in years
     */
    public void setAge(Integer age) {
        this.age = age;
    }

    /**
     * an awesome name!
     */
    public String getName() {
        return name;
    }
    /**
     * an awesome name!
     */
    public void setName(String name) {
        this.name = name;
    }



    public static class Builder {

        private String name;
        private Integer age;
        private String name;

        public Builder() {
        }

        public Builder setName(String name) {
            this.name = name;
            return this;
        }

        /**
         * age in years
         */
        public Builder setAge(Integer age) {
            this.age = age;
            return this;
        }

        /**
         * an awesome name!
         */
        public Builder setName(String name) {
            this.name = name;
            return this;
        }


        public Person build() {
            return new Person(name, age, name);
        }

    }
}
```

## Your Environment and Setup

* **graphql-java-codegen**: 1.7.0
* **Build tool**: Maven
* **Java tool**: openjdk 11.0.6 2020-01-14
* **Mapping Config**: It won't tell you much. It's just the paths config

```xml
<configuration>
  <graphqlSchemaPaths>
    <graphqlSchemaPath>${graphql.schema.path}</graphqlSchemaPath>
  </graphqlSchemaPaths>
  <outputDir>${graphql-codegen-maven-plugin.outputDir}</outputDir>
  <packageName>${server.packageName}</packageName>
</configuration>
```

## Additional context
It works with 1.6.0

kobylynskyi:
Fixes #128

### DIFF
--- a/src/main/java/com/kobylynskyi/graphql/codegen/mapper/TypeDefinitionToDataModelMapper.java
+++ b/src/main/java/com/kobylynskyi/graphql/codegen/mapper/TypeDefinitionToDataModelMapper.java
@@ -93,8 +93,33 @@ public class TypeDefinitionToDataModelMapper {
         getInterfacesOfType(typeDefinition, document).stream()
                 .map(i -> FieldDefinitionToParameterMapper.mapFields(mappingConfig, i.getFieldDefinitions(), i.getName()))
                 .flatMap(Collection::stream)
-                .forEach(p -> allParameters.put(p.getName(), p));
+                .forEach(paramDef -> {
+                    if (allParameters.containsKey(paramDef.getName())) {
+                        ParameterDefinition existingParamDef = allParameters.get(paramDef.getName());
+                        allParameters.put(paramDef.getName(), merge(existingParamDef, paramDef));
+                    } else {
+                        allParameters.put(paramDef.getName(), paramDef);
+                    }
+                });
         return allParameters.values();
+    }
+
+    /**
+     * Merge parameter definition data from the type and interface
+     * Annotations from the type have higher precedence
+     *
+     * @param typeDef      Definition of the same parameter from the type
+     * @param interfaceDef Definition of the same parameter from the interface
+     * @return merged parameter definition
+     */
+    private static ParameterDefinition merge(ParameterDefinition typeDef, ParameterDefinition interfaceDef) {
+        if (Utils.isEmpty(typeDef.getAnnotations())) {
+            typeDef.setAnnotations(interfaceDef.getAnnotations());
+        }
+        if (Utils.isEmpty(typeDef.getJavaDoc())) {
+            typeDef.setJavaDoc(interfaceDef.getJavaDoc());
+        }
+        return typeDef;
     }
 
     /**
@@ -120,7 +145,8 @@ public class TypeDefinitionToDataModelMapper {
         getInterfacesOfType(typeDefinition, document).stream()
                 .map(i -> FieldDefinitionToParameterMapper.mapProjectionFields(mappingConfig, i.getFieldDefinitions(), typeNames))
                 .flatMap(Collection::stream)
-                .forEach(p -> allParameters.put(p.getName(), p));
+                .filter(paramDef -> !allParameters.containsKey(paramDef.getName()))
+                .forEach(paramDef -> allParameters.put(paramDef.getName(), paramDef));
         return allParameters.values();
     }
 
--- a/src/main/java/com/kobylynskyi/graphql/codegen/mapper/TypeDefinitionToDataModelMapper.java
+++ b/src/main/java/com/kobylynskyi/graphql/codegen/mapper/TypeDefinitionToDataModelMapper.java
@@ -93,14 +93,7 @@ public class TypeDefinitionToDataModelMapper {
         getInterfacesOfType(typeDefinition, document).stream()
                 .map(i -> FieldDefinitionToParameterMapper.mapFields(mappingConfig, i.getFieldDefinitions(), i.getName()))
                 .flatMap(Collection::stream)
-                .forEach(paramDef -> {
-                    if (allParameters.containsKey(paramDef.getName())) {
-                        ParameterDefinition existingParamDef = allParameters.get(paramDef.getName());
-                        allParameters.put(paramDef.getName(), merge(existingParamDef, paramDef));
-                    } else {
-                        allParameters.put(paramDef.getName(), paramDef);
-                    }
-                });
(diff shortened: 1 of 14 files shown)


src/main/java/com/kobylynskyi/graphql/codegen/mapper/TypeDefinitionToDataModelMapper.java
```diff
@@ -80,22 +80,21 @@
...
+                                                             ExtendedObjectTypeDefinition typeDefinition,
+                                                             ExtendedDocument document) {
+        // using the map to exclude duplicate fields from the type and interfaces
+        Map<String, ParameterDefinition> allParameters = new LinkedHashMap<>();
+
+        // includes parameters from the base definition and extensions
+        FieldDefinitionToParameterMapper.mapFields(mappingConfig, typeDefinition.getFieldDefinitions(), typeDefinition.getName())
+                .forEach(p -> allParameters.put(p.getName(), p));
+        // includes parameters from the interface
+        getInterfacesOfType(typeDefinition, document).stream()
```
joffrey-bion, line 93: Why do we need to get fields from the interface in the first place?
By definition, for any implemented interface, all fields should be declared in the type itself as well.
Is it to get metadata?

kobylynskyi: Sorry, I didn’t get your point.
We are getting all fields fields from the type first. And then adding fields from the interface

joffrey-bion: Yes, that's why we need to read the interface's fields as well, indeed.
However, we could imagine a situation where the doc is overridden in the type (or not even present in the interface), and in that case we need to take the doc from the type's field, not the interface's one

kobylynskyi: I've modified the logic to take JavaDoc and annotations from the type. 
And if something is absent then it will be taken from the interface field.

src/main/java/com/kobylynskyi/graphql/codegen/mapper/TypeDefinitionToDataModelMapper.java
```diff
@@ -80,22 +80,46 @@
...
                 .map(i -> FieldDefinitionToParameterMapper.mapFields(mappingConfig, i.getFieldDefinitions(), i.getName()))
                 .flatMap(Collection::stream)
-                .collect(Collectors.toList());
+                .forEach(paramDef -> {
+                    if (allParameters.containsKey(paramDef.getName())) {
+                        ParameterDefinition existingParamDef = allParameters.get(paramDef.getName());
+                        allParameters.put(paramDef.getName(), merge(existingParamDef, paramDef));
+                    } else {
+                        allParameters.put(paramDef.getName(), paramDef);
+                    }
```
joffrey-bion, line 102: This is a minor comment, but the whole `if` block can be replaced by:
```java
allParameters.merge(paramDef.getName(), paramDef, ::merge)
```
Or something along those lines.

src/main/java/com/kobylynskyi/graphql/codegen/mapper/TypeDefinitionToDataModelMapper.java
joffrey-bion: Now I'm wondering, do we really want to have this at all? I mean, to generate the Java code for the interface, we use the interface fields and docs. And to generate the Java code for the implementing class, we don't need to look at the interface fields at all, do we?
Javadoc would naturally be inherited by overriding methods (we don't even need to do this at codegen level). And annotations are not inherited, but do we want that?

kobylynskyi: Could you please create a separate issue with your suggestion? 
We can discuss it there.
Thanks!

src/main/java/com/kobylynskyi/graphql/codegen/mapper/TypeDefinitionToDataModelMapper.java
kobylynskyi: Thank you so much. I never used this method before.

src/main/java/com/kobylynskyi/graphql/codegen/mapper/TypeDefinitionToDataModelMapper.java
joffrey-bion: Sure I can do that, I just felt like it was another way to fix the issue at hand, not a separate issue.
But you're right, we can merge this and fix the bug right away.

Issue #145, opened 2020-05-14
Methods returning interface cause compilation error (Version: [1.7.3])


ClaudiuGuja:
The issue:
    Code generated from the following SDL generates the interface Foo with `Collection<Bar> getBars()` method and the implementation Foo1 with `Collection<Bar1> getBars()` method. Since the interface method is not overwritten, a compilation error takes place.

Proposed solution:
Perhaps the interface can be generated like this: `Collection<? extends Bar> getBars()` if the return type of the method would be an interface

The SDL:
```
interface Foo {
    id: ID!
    bars: [Bar]
}

interface Bar {
    id: ID!
}

type Foo1 implements Foo{
    id: ID!
    bars: [Bar1]
}

type Bar1 implements Bar{
    id: ID!
}

```

kobylynskyi:
@ClaudiuGuja 
Fixed in [1.7.4](https://github.com/kobylynskyi/graphql-java-codegen/releases/tag/v1.7.4)
Please create another issue if you see any problems.

PR #147, opened 2020-05-15
Use upper bounded wildcard for interface methods returning list #145

kobylynskyi:


kobylynskyi:
@joffrey-bion could you please review this PR when you have time? Thanks.

joffrey-bion:
Hi @kobylynskyi , sorry for the delay in review, I actually didn't have much time to take a look these days, and then I forgot about it.

This looks very nice!

I'm just wondering whether some of test files were added on purpose or not. The `expected-classes` make sense, but I'm not sure you meant to check in the ones like `src/test/java/com/kobylynskyi/graphql/codegen/model/Foo1.java.txt`. Is this intended?

kobylynskyi:
@joffrey-bion, thanks for your review!
Yes, it is intended. There is a test which generates classes with such names (copies from the corresponding issue

joffrey-bion:
@kobylynskyi but if it's generated then it shouldn't be checked in the repo, should it?

kobylynskyi:
@joffrey-bion, I've removed the classes you mentioned. Thank you for pointing this out!

PR #152, opened 2020-05-24
feat: Support for enum in unions

danilo-barboza:
First of all great job on this library, works great and would really help a lot on having schema first applications in Java!

This PR is to solve one issue I had while testing it in one of my use cases for graphql where there were a schema with a `union` that has enums in it.

As this is a valid schema and in Java enums can implement interfaces, it would make sense to also support that in the code generation.

I added a new test case for it where you can clearly see the use case and the code generation.

Please let me know if there is anything else you think would be necessary.


kobylynskyi:
Released in [1.8.0](https://github.com/kobylynskyi/graphql-java-codegen/releases/tag/v1.8.0)

danilo-barboza:
Awesome! Thanks @kobylynskyi!

PR #156, opened 2020-05-27
Externalize codegen options to a separate file

kobylynskyi:


docs/codegen-options.md
```diff
@@ -0,0 +1,77 @@
...
+| `requestSuffix`                                 | String                                             | Request                                       | Sets the suffix for `Request` classes. |
+| `responseProjectionSuffix`                      | String                                             | ResponseProjection                            | Sets the suffix for `ResponseProjection` classes. |
+| `parametrizedInputSuffix`                       | String                                             | ParametrizedInput                             | Sets the suffix for `ParametrizedInput` classes. |
+| `parentInterfaces`                              | *See [parentInterfaces](#option-parentinterfaces)* | Empty                                         | Block to define parent interfaces for generated interfaces (query / mutation / subscription / type resolver) |
+
+
+### Option `graphqlSchemas`
+
+When exact paths to GraphQL schemas are too cumbersome to provide in the `graphqlSchemaPaths`, use the `<graphqlSchemas></graphqlSchemas>` block.
```
joffrey-bion, line 37: This syntax is maven-specific, there used to be a couple differences like this between the 2 plugin docs (maybe it was overridden already because I don't see the differences anymore).

I like that you extracted the options, it avoids duplication, I think it's worth giving up on the syntax details of each build tool. In this case, though, I think we should avoid the XML syntax and maybe mention just "`graphqlSchemas` block" like this, which would sort of make sense for both plugins.

docs/codegen-options.md
```diff
@@ -0,0 +1,77 @@
...
+| `fieldsWithoutResolvers`                        | Set(String)                                        | Empty                                         | Fields that DO NOT require Resolvers should be defined here in format: `TypeName.fieldName` or `TypeName`. Can be used in conjunction with `generateExtensionFieldsResolvers` option. |
+| `generateParameterizedFieldsResolvers`          | Boolean                                            | True                                          | If true, then generate separate `Resolver` interface for parametrized fields. If false, then add field to the type definition and ignore field parameters. |
+| `generateExtensionFieldsResolvers`              | Boolean                                            | False                                         | Specifies whether all fields in extensions (<code>extend type</code> and <code>extend interface</code>) should be present in Resolver interface instead of the type class itself. |
```
joffrey-bion, line 26: Maybe we can simplify `<code>` using backticks instead. That's totally unnecessary nitpicking (I'm sure I have undiagnosed OCD) 😄

docs/codegen-options.md
kobylynskyi: That's a good catch. I've fixed it as per your suggestion.
Thanks, @joffrey-bion!

docs/codegen-options.md
kobylynskyi: Thanks!

kobylynskyi:
@joffrey-bion thanks! I appreciate your review.

Issue #154, opened 2020-05-26
Missing simple request serialization


lmartellotto:
Hello,

**Is your feature request related to a problem? Please describe.**
We use this project to test our GraphQL API with RestAssured. Our API exposes two endpoints: GET and POST, as mentionned here: https://graphql.org/learn/serving-over-http/.

We generate requests as follow:
```java
ItemByIdQueryRequest query = new ItemByIdQueryRequest.Builder()
                                                          .setId(1)
                                                          .build();
        
ItemResponseProjection fields = new ItemResponseProjection()
                                                          .name()
                                                          .tags();
GraphQLRequest request = new GraphQLRequest(query, fields);
```

... which generates GraphQL following JSON query thanks to `request.toString():`
```
{"query":"query { itemById(id: 1){ name tags } }"}
```
This JSON can be sent to the POST endpoint, it works fine. But not for the GET endpoint! Because the GET endpoint is waiting for "query" parameter where the value is not a JSON query, but the GraphQL query itself. So, we need to have only:

```
query { itemById(id: 1){ name tags } }
```

**Describe the solution you'd like**
Maybe just a new getter in the `GraphQLRequest` class, which returns the GraphQL query without JSON encapsulation. 

What about a `toString()` method overload, like:

```java
    @Override
    public String toString() {
        return toString(true);
    }

    public String toString(final boolean jsonEncapsulation) {
        // TODO
    }
```

All the serialization work is already done in the `GraphQLRequestSerializer.serialize()`; it just needs to not call the last method `buildJsonQuery()` when we don't need JSON encapsulation.

If you are ok with that, I can submit a PR :)

Thanks for your time,
Laura

kobylynskyi:
@lmartellotto This sounds good to me!

joffrey-bion:
That makes total sense for me as well. I would even suggest that we refrain from using the name `toString()` for the JSON transformation, but maybe that's a backward compatibility issue :/

kobylynskyi:
@joffrey-bion actually no, it is not a backward-compatibility issue.
`toString()` method of the generated class is being called only from `GraphQLRequestSerializer`.

kobylynskyi:
@lmartellotto I suggest having following methods in `GraphQLRequest` class:
```java

    /**
     * @deprecated Not intended for use and will be removed in the next version.
     * Please use one of: {@link #toJsonRequest} or {@link #toRequest}
     */
    @Override
    @Deprecated
[...]

PR #157, opened 2020-05-27
Support of HTTP Query String serialization #154

lmartellotto:
* Support of HTTP Query String serialization #154

* Add junit-jupiter-params as test dependency + parameterized tests for GraphQLRequestSerializer

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -25,6 +25,7 @@ dependencies {
     annotationProcessor "org.projectlombok:lombok:1.18.12"
 
     testImplementation "org.junit.jupiter:junit-jupiter-api:5.5.1"
+    testImplementation "org.junit.jupiter:junit-jupiter-params:5.5.1"
     testRuntimeOnly "org.junit.jupiter:junit-jupiter-engine:5.5.1"
     testImplementation 'org.hamcrest:java-hamcrest:2.0.0.0'
 }
--- a/src/main/java/com/kobylynskyi/graphql/codegen/model/graphql/GraphQLRequest.java
+++ b/src/main/java/com/kobylynskyi/graphql/codegen/model/graphql/GraphQLRequest.java
@@ -25,8 +25,32 @@ public class GraphQLRequest {
         return responseProjection;
     }
 
+    /**
+     * @deprecated Not intended for use and will be removed in the next version.
+     * Please use one of: {@link #toHttpJsonBody} or {@link #toQueryString}
+     */
     @Override
+    @Deprecated
     public String toString() {
-        return GraphQLRequestSerializer.serialize(this);
+        return toHttpJsonBody();
+    }
+
+    /**
+     * Serializes GraphQL request to be used as HTTP JSON body
+     * according to https://graphql.org/learn/serving-over-http specifications
+     * 
+     * @return the serialized request
+     */
+    public String toHttpJsonBody() {
+        return GraphQLRequestSerializer.toHttpJsonBody(this);
+    }
+
+    /**
+     * Serializes GraphQL request as raw query string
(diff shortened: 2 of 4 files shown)


build.gradle
```diff
@@ -25,6 +25,7 @@ dependencies {
     annotationProcessor "org.projectlombok:lombok:1.18.12"
 
     testImplementation "org.junit.jupiter:junit-jupiter-api:5.5.1"
+    testImplementation "org.junit.jupiter:junit-jupiter-params:5.5.1"
```
lmartellotto, line 28: I add this dependency to be able to write parameterized tests with JUnit 5. If you don't agree with that, I can remove and duplicate all tests in GraphQLRequestSerializerTest.

src/main/java/com/kobylynskyi/graphql/codegen/model/graphql/GraphQLRequest.java
```diff
@@ -25,8 +25,33 @@ public GraphQLResponseProjection getResponseProjection() {
...
+        return GraphQLRequestSerializer.toHttpJsonBody(this);
+    }
+
+    /**
+     * Serializes GraphQL request to be used as HTTP query string 
+     * according to https://graphql.org/learn/serving-over-http specifications
+     * 
+     * @return the serialized request
+     */
+    public String toHttpQueryString() {
```
lmartellotto, line 54: @kobylynskyi @joffrey-bion Finally, are you ok with `toHttpJsonBody` and `toHttpQueryString` ? If we mention HTTP for the `toHttpJsonBody` method, maybe it will be more consistent to mention HTTP for the `toHttpQueryString` method too.

If you don't agree with that, I can rename the method and all mentions to it, no problem :)

src/main/java/com/kobylynskyi/graphql/codegen/model/graphql/GraphQLRequestSerializer.java
```diff
@@ -13,7 +13,38 @@
...
+     */
+    public static String toHttpQueryString(GraphQLRequest graphQLRequest) {
+        return buildQuery(graphQLRequest, false);
+    }
+
+    /**
+     * @deprecated Not intended for use and will be move to private method in the next version.
+     * Please use one of: {@link #toHttpJsonBody} or {@link #toHttpQueryString}
+     */
+    @Deprecated
```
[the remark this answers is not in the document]
kobylynskyi: Never mind. We can leave it `@Deprecated` for now and remove it in 2.0.0.

src/main/java/com/kobylynskyi/graphql/codegen/model/graphql/GraphQLRequest.java
joffrey-bion: Hi Laura, sorry for the delay. I was actually envisioning `toQueryString()` to be the raw GraphQL query string, not the HTTP one, so that people can build upon it. Actually, `toHttpQueryString()` would be a good *addition* to that (which would URL-encode the query returned by `toQueryString()` I guess), but I believe we should keep the raw GraphQL query available to users for extension.
What do you guys think?

lmartellotto: @joffrey-bion I totally agree with you! Indeed, I didn't think about the raw GraphQL query string. It makes sense. So, I will rename it.

About an additionnal `toHttpQueryString()` or `toUrlQueryString()` (both name are fine to me too), it would be nice. I can add it but maybe in other issue, I have no more time for this week 😕 

Thanks Joffrey!

PR #161, opened 2020-05-29
Generate Response class having response data with exact type #159

### LINKED ISSUE #159: Mapping and accessing "partial" responses
I'm having trouble wrapping my head around the mapping between whatever GraphQLResult returns after a successful query and the generated classes.

I'm executing a mutation and fetch only a single field "access_token" from the response by setting the appropriate projection. The autogenerated response object "LoginUserResponse" has another field "user", which I'm not interested in.

```java
public void login() throws IOException, GraphQLException {
    LoginUserInput input = new LoginUserInput();
    input.setUsername(this.username);
    input.setPassword(this.password);

    LoginUserMutationRequest query = new LoginUserMutationRequest();
    query.setUser_data(input);

    // Only fetch field access_token, ignore all other fields
    LoginUserResponseResponseProjection projection = new LoginUserResponseResponseProjection().access_token();

    GraphQLResult<Map<String, LoginUserResponse>> result = this.queryEntity(query, projection);
    System.out.println(result.getData().get(query.getOperationName()).getAccess_token());
}

public <Q extends GraphQLOperationRequest, P extends GraphQLResponseProjection, R> GraphQLResult<Map<String, R>> queryEntity(Q query, P projection)  throws GraphQLException {
    GraphQLRequest request = new GraphQLRequest(query, projection);
    GraphQLResult<Map<String, R>> result = restTemplate.exchange(URI.create(this.endpoint + "/***/api/graphql"),
        HttpMethod.POST,
        new HttpEntity<String>(request.toHttpJsonBody(), createHeaders(request)),
        new ParameterizedTypeReference<GraphQLResult<Map<String, R>>>() {})
        .getBody();
    if (result.hasErrors()) {
        throw new GraphQLException(result.getErrors());
    }
    return result;
}
```

This leads to this error:

`java.lang.ClassCastException: class java.util.LinkedHashMap cannot be cast to class com.***.***.model.LoginUserResponse (java.util.LinkedHashMap is in module java.base of loader 'bootstrap'; com.***.***.model.LoginUserResponse is in unnamed module of loader 'app')`

Which makes sense, as the class LoginResponse would expect the field 'user' to be not null, but I'm not fetching this field, so constructing an instance of LoginResponse is impossible.

Could anyone give me a hint on how to properly access the fields I fetched? I can see that there are Mapper definitions in the examples, but I'm a little confused regarding the differences between ProductTO and Product, as I can not find two different classes for an entity in my autogenerated code.

Any help very much appreciated!

kobylynskyi:
@philipp-schmidt could you please check if that's what you want?

---
Resolves #159


philipp-schmidt:
Hello @kobylynskyi 
Looking at the changes (primarily at the updated examples), this looks good to me. However, I don't have the time right now to test this locally before there is a release candidate (I don't have much experience with maven plugins and local maven repository etc. as described in Contributing.md).
Any chance these changes will make it into a release anytime soon?

Thanks again for the time and effort you put into this!

PR #168, opened 2020-06-05
Model classes should implement java.io.Serializable #167

### LINKED ISSUE #167: Generated model classes should implement java.io.Serializable
I want to handle model objects transparently, could you add the ability to implement the java.io.Serializable interface?

kobylynskyi:
Resolves #167


kobylynskyi:
@joffrey-bion could you please take a look? Thanks!

src/main/java/com/kobylynskyi/graphql/codegen/utils/SerialVersionUIDGenerator.java
```diff
@@ -0,0 +1,16 @@
...
+
+public class SerialVersionUIDGenerator {
+
+    /**
+     * Generate random long
+     *
+     * @return random long
+     */
+    public long randomLong() {
+        return ThreadLocalRandom.current().nextLong();
```
joffrey-bion, line 13: I have concerns about this.

AFAIU, a different serial version UID will be generated each time the code generation is triggered.
This means that even if the schema doesn't change, all classes will get a new UID and thus be declared "incompatible" (that's the meaning of the serial UID).

I think we should rely on the compiler-generated one without attempting to generate our own. It may be safer

src/test/java/com/kobylynskyi/graphql/codegen/TestUtils.java
```diff
@@ -23,4 +26,17 @@ public static void assertSameTrimmedContent(File expected, File file) throws IOE
...
+
+    public static MappingConfig initMappingConfig() {
+        MappingConfig mappingConfig = new MappingConfig();
+        mappingConfig.setSerialVersionUIDGenerator(getSerialVersionUIDGenerator());
+        return mappingConfig;
+    }
+
+    public static SerialVersionUIDGenerator getSerialVersionUIDGenerator() {
+        SerialVersionUIDGenerator serialVersionUIDGenerator = Mockito.mock(SerialVersionUIDGenerator.class);
+        Mockito.when(serialVersionUIDGenerator.randomLong()).thenReturn(123456L);
```
joffrey-bion, line 38: This mock actually shows the problem. Serial version UID should be deterministic and not require mocking

src/main/java/com/kobylynskyi/graphql/codegen/utils/SerialVersionUIDGenerator.java
kobylynskyi: Thanks for your feedback. I also had concerns whether to define an explicit serialVersionUID or not.
But the following phrase overweight me:
“ However, it is strongly recommended that all serializable classes explicitly declare serialVersionUID values, since the default serialVersionUID computation is highly sensitive to class details that may vary depending on compiler implementations, and can thus result in unexpected InvalidClassExceptions during deserialization.”

src/test/java/com/kobylynskyi/graphql/codegen/TestUtils.java
kobylynskyi: The idea of hardcoding serialVersionUID here was because we need to compare generated class with the expected one.
But this mocking can be removed since explicit serialVersionUID is not recommended, right?

src/main/java/com/kobylynskyi/graphql/codegen/utils/SerialVersionUIDGenerator.java
joffrey-bion: @kobylynskyi what the sentence means essentially is that if you make a backwards *compatible* change in the class, due to the sensitivity of the generation algorithm, you may get a different version UID anyway. This will trigger `InvalidClassException`s at deserialization time even if the deserialization would in fact have worked.

But in our case, generating it randomly is even worse than that: we will *always* generate incompatible versions even if no changes are made in the class at all. So classes generated by 2 different builds will always be incompatible and trigger `InvalidClassException`s.
So I guess for the use case of generated code, not generating a serial version UID is the lesser evil.

We could try to attempt generating a serial version UID based on the actual schema and details of the class generated (so that the same class yields the same version ID), but that would basically be reimplementing the auto-generation of the serialization runtime, so I guess we can assume they already did a good job.

The only superior approach would be manual attribution of ID (which is the recommended way for classes you write manually and own), but it doesn't suit our use case.

kobylynskyi: Regarding custom directive - that's a nice idea! I think we can consider this in the future.

PR #179, opened 2020-06-17
Fix a case when List NonNullType has a default value #177

### LINKED ISSUE #177: Code generation using gradle plugin fails for NonNullType of List with empty list as default value (Version: graphql-java-codegen 1.7.3)
## Issue Description

Code generation using gradle plugin from graphql schema fails with this message:
`Unexpected array default value for non-list type`

Here is a snippet of the schema, I actually identified the part that fails:
```
type Query {
  "Select error events"
  select_error(
    "Maximum number of objects to return"
    first: Int! = 1000, 
    "List of possible options for the group by"
    group_by: [bus_error_groupby_enum!]! = []
  ): [error_activity]
```

the part that causes the issue is this:
`group_by: [bus_error_groupby_enum!]! = []`
 changing it to this, gets rid of the issue:
`group_by: [bus_error_groupby_enum!] = []`

I tried the code to understand where it came from and apparently it is in 
src/main/java/com/kobylynskyi/graphql/codegen/mapper/DefaultValueMapper.java

```
  private static String mapArray(MappingContext mappingContext, ArrayValue defaultValue, Type<?> graphQLType) {
        if (!(graphQLType instanceof ListType)) {
            throw new IllegalArgumentException("Unexpected array default value for non-list type");
        }
```
the defaultValue is a list while the type is actually not a ListType but a NonNullType

## Steps to Reproduce
Add this to your schema and try to generate the code
`group_by: [bus_error_groupby_enum!]! = []`

## Expected Result
I would expect a NonNullType of List to accept a default value of type List

## Actual Result
Code generation using gradle plugin from graphql schema fails with this message:
`Unexpected array default value for non-list type`

## Your Environment and Setup

* **graphql-java-codegen**: *1.7.3*
* **Build tool**: *Gradle*
* **Java tool**: *AdoptOpenJdk 11.0.6.10-hotspot*
* **Mapping Config**: 

```
graphqlCodegen {
    graphqlSchemaPaths = ["$projectDir/generated_schema.graphqls".toString()]
    outputDir = new File("$buildDir/generated")
    generateApis = true
    generateToString = true
    modelValidationAnnotation = ""
    generateRequests = true
}

```

kobylynskyi:
Fixes #177

### DIFF
--- a/src/main/java/com/kobylynskyi/graphql/codegen/mapper/DefaultValueMapper.java
+++ b/src/main/java/com/kobylynskyi/graphql/codegen/mapper/DefaultValueMapper.java
@@ -31,7 +31,7 @@ public class DefaultValueMapper {
             return mapObject((ObjectValue) defaultValue);
         }
         if (defaultValue instanceof ArrayValue) {
-            return mapArray(mappingContext, (ArrayValue) defaultValue, graphQLType);
+            return mapArray(mappingContext, graphQLType, (ArrayValue) defaultValue);
         }
         // no default value, or not a known type
         return null;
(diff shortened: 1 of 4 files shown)


lorenzleutgeb:
LGTM

kobylynskyi:
@lorenzleutgeb thanks for your review! 
Merging this now. All future comments/notes can be addressed later if required.

PR #186, opened 2020-06-18
Use java.util.List for GQL lists

### LINKED ISSUE #185: Use java.util.List for lists
TL;DR: I think that the implementation of this generator misinterprets the semantics of lists in GraphQL. It generates `Collection<T>` where it should generate `List<T>`.

There's mismatch in what classes generated by the library *are expected to provide* (lists) and what they *actually provide* (collections).

### Steps to Reproduce

Generate code for

```graphql
type Foo { bar: [String] }
```

### Expected Result

The Java equivalent of `bar` should have type `List<String>`, and this should of course generalize.

### Actual Result

The Java equivalent of `bar` has type `Collection<String>`.

### Semantics of Lists in GraphQL

[The GraphQL specification (draft)](https://spec.graphql.org/draft/#sec-Type-System.List) considers a list to be
> List values are serialized as **ordered lists**, where each item in the list is serialized as per the item type.

(emphasis mine) which clearly references ordering of elements as a characteristic property of values of the list type.

Also there is [other tooling out there, like Hasura](https://hasura.io/docs/1.0/graphql/manual/queries/sorting.html#sorting-objects) which clearly rely on the property of lists being ordered.

### Semantics of Lists in this generator

The `Collection` type in Java, however, does not guarantee ordering of elements (consider `HashSet` as a counterexample).

lorenzleutgeb:
Resolves #185

This is a breaking change, and should hopefully make it into version 2.

### DIFF
--- a/plugins/gradle/example-client/src/main/java/io/github/kobylynskyi/order/external/ProductServiceGraphQLClient.java
+++ b/plugins/gradle/example-client/src/main/java/io/github/kobylynskyi/order/external/ProductServiceGraphQLClient.java
@@ -25,7 +25,6 @@ import org.springframework.web.client.RestTemplate;
 
 import java.math.BigDecimal;
 import java.net.URI;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.stream.Collectors;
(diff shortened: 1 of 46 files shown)


lorenzleutgeb:
Cool. Do you have a timeline for releasing 2?

kobylynskyi:
I am planning to do it today (once other open pull requests will be merged)

PR #236, opened 2020-07-17
Custom types for async Query/Mutation generation #228

### LINKED ISSUE #228: Support custom type for Query and Mutation in async API generation
As describe in the wiki, `generateAsyncApi` if is true return `java.util.concurrent.CompletableFuture` or `subscriptionReturnType`. That a bit equivocal cause you can understand it as it return `subscriptionReturnType` if is set (but that never the case for Query and Mutation as it always return `CompletableFuture`). 

But in some project with Spring Webflux, it can be useful to have a Query interface that returns `Mono` or `Flux` rather than `CompletableFuture` but atm, we can't.

To not break already implemented projects, add a new parameter in Codegen Options that define a custom return type for Query and Mutation interfaces. 

If there is no value or empty value and `generateAsyncApi`is true, the plugin use `CompletableFuture` as default.

Wrap all `CompletableFuture` into `Mono` or `Flux` but it's kind of annoying

**Additional context**
Tested with :
- Spring Webflux: 5.2.x
- graphql-java-codegen: 2.2.0
- Build tool: Maven

kobylynskyi:
@joffrey-bion @sriio can you guys please take a look?

---

Resolves #228


kobylynskyi:
I will merge it now, but will still address the comments if there will be any. Thanks!

PR #250, opened 2020-07-25
Handle java reserved words in enum #249

### LINKED ISSUE #249: Changed enum capitalization when using java reserved words (Version: 2.4.0)
## Issue Description

Using reserved java words as enum names results in capitalization which is then used in the query. This results in a rejected request (by AWS Appsync) because the enum is not defined in the schema.

## Steps to Reproduce

Let's assume the following graphql schema:

```graphql
type Mutation {
    updateTest(message: TestEnum): String
}

enum TestEnum {
    long,
    short,
    high,
    low
}
```

and generate the model and client from it. We use it like this:

```java
public static void main(String[] args) {
    UpdateTestMutationRequest mutReq = new UpdateTestMutationRequest();
    mutReq.setMessage(TestEnumTO.Long);
    GraphQLRequest req = new GraphQLRequest(mutReq);
    String json = GraphQLRequestSerializer.toHttpJsonBody(req);
    System.out.println(json);
}
```
The printed line is the query later sent to the server. 

## Expected Result
```
{"query":"mutation { updateTest(message: long) }"}
```

## Actual Result
```
{"query":"mutation { updateTest(message: Long) }"}
```
Since the query is generated [by the serializer using #toString()](https://github.com/kobylynskyi/graphql-java-codegen/blob/03e6b5cc12a5a729676222f49dae21d7ecbe5c9a/src/main/java/com/kobylynskyi/graphql/codegen/model/graphql/GraphQLRequestSerializer.java#L98) the capitalized name is used.

## Your Environment and Setup

* **graphql-java-codegen**: 2.4.0
* **Build tool**: Maven
* **Java tool**: openjdk version "1.8.0_252"; OpenJDK Runtime Environment (build 1.8.0_252-8u252-b09-1ubuntu1-b09)

* **Mapping Config**:

```xml
<configuration>
    <graphqlSchemas>
        <rootDir>${project.basedir}/schema</rootDir>
    </graphqlSchemas>
    <outputDir>${project.build.directory}/generated-sources/graphql</outputDir>
    <packageName>test.model</packageName>
    <modelNameSuffix>TO</modelNameSuffix>
    <generateClient>true</generateClient>
</configuration>
```

## Additional context

- The problem seems similar to https://github.com/kobylynskyi/graphql-java-codegen/issues/237.
- I only looked at the client side when building the query. There might be a problem when receiving a response as well.
- I temporarily worked around this problem using https://github.com/Nithanim/graphql-java-codegen/commit/b8f5c0085e4c41e63f3764f7c50492d7b702b2ad which seemed to work flawlessly.
- Thank you for the awesome generator!
- [Minimal maven test project](https://github.com/kobylynskyi/graphql-java-codegen/files/4976562/minimal.zip)

kobylynskyi:
Fixes #249 
@Nithanim could you please take a look?

### DIFF
--- a/src/main/java/com/kobylynskyi/graphql/codegen/mapper/EnumDefinitionToDataModelMapper.java
+++ b/src/main/java/com/kobylynskyi/graphql/codegen/mapper/EnumDefinitionToDataModelMapper.java
@@ -74,6 +74,7 @@ public class EnumDefinitionToDataModelMapper {
         return enumValueDefinitions.stream()
                 .map(f -> new EnumValueDefinition(
                         MapperUtils.capitalizeIfRestricted(f.getName()),
+                        f.getName(),
                         getJavaDoc(f.getComments()),
                         isDeprecated(f)))
                 .collect(Collectors.toList());
--- a/src/main/java/com/kobylynskyi/graphql/codegen/mapper/MapperUtils.java
+++ b/src/main/java/com/kobylynskyi/graphql/codegen/mapper/MapperUtils.java
@@ -27,8 +27,8 @@ public class MapperUtils {
 
     private static final Set<String> JAVA_RESTRICTED_KEYWORDS = new HashSet<>(Arrays.asList(
             "abstract", "assert", "boolean", "break", "byte", "case", "catch", "char", "class", "const", "continue",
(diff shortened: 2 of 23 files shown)


Nithanim:
You are faster fixing than I am done sleeping. Thank you very much for your fast response!

PR #274, opened 2020-08-10
Remove generateAsyncApi, rename apiReturnType, apiReturnListType #269

### LINKED ISSUE #269: Support for Optional when returning nullable types
Thank you for the library, it's great! 👏 

Because the query resolvers return the nullable object (without wrapping it in Optional) it can feel outdated for people heavily using Java8+ features.

Currently when you have a query that has a nullable return type like this:

```graphql
type Query {
    user(id: ID!): User
}
```

the generated code is:

```java
@Override
public User user(final String id) {
    return null;
}
```

Would be nice to have a flag to enable/disable behaviour that allows the generation for nullable return types to use Optional instead which can make it easier to write neater code in the resolver.

```java
@Override
public User user(final String id) {
    return Optional.empty();
}
```

None

**Additional context**
None

kobylynskyi:


jxnu-liguobin:
Excuse me, what is the purpose of removing this? Reduce the number of options?

kobylynskyi:
If you want it use async APIs then just supply apiReturnType / apiReturnListType

Issue #291, opened 2020-08-17
If java code with default values is generated, type mismatch may occur.


jxnu-liguobin:
**Is your feature request related to a problem? Please describe.**
graphql schema has a field like `interval: Long = 3600000`
will create java code like `private java.lang.Long interval = 3600000;` but `int` can not assign to `Long` in java
there are two ways resolver it.
1. generate code: `private java.lang.Long interval = 3600000L;`
2. generate code: `private long interval = 3600000;` then need add `types.put ("Long", "long")` in customTypesMapping
and can not use `!=` 
```java
    if (interval != null) {//error,  != cannot applied to long and null
            joiner.add("interval: " + GraphQLRequestSerializer.getEntry(interval));
     }
```
At now i use `types.put("Long", "java.lang.Long")`, so will generate a error java code.
And can not append `L` in graphql. At present, in order to modify the minimum code, I am ready to use the first way.
maybe support the first way is easy, i think `graphql-java-codegen` maybe need support  wrap default value.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

kobylynskyi:
@jxnu-liguobin thanks for reporting this issue

kobylynskyi:
Will be released as part of 3.0.1 by the end of the week

PR #292, opened 2020-08-17
add suffix after the default value of the base type(long)

jxnu-liguobin:
#291


src/main/java/com/kobylynskyi/graphql/codegen/mapper/ValueMapper.java
```diff
@@ -65,7 +67,28 @@ private static String mapBoolean(BooleanValue value) {
...
     }
 
-    private static String mapInt(IntValue value) {
+    private static String mapInt(MappingContext mappingContext, IntValue value, Type<?> graphQLType) {
+        //default java basic type is `int`. so, default value like 123 that must wrap or append suffix `L` when it be defined as `int` in graphql schema.
+        //`int` cannot assign to `Long`, also `double` cannot assign to `Float`, but graphql Float default mapping is Double in java, so, not modify `mapFloat`.
+        if (graphQLType != null) {
+            List<Field> fields = Arrays.stream(graphQLType.getClass().getDeclaredFields()).collect(Collectors.toList());
+            String customType = mappingContext.getCustomTypesMapping().get("Long");
+            for (Field f : fields) {
```
kobylynskyi, line 76: @jxnu-liguobin 
1. Could you please explain what are you trying to get from `graphQLType fields`?
2. In `customTypesMapping` you are checking for a key `Long`, but someone can have scalar called `myLong` and a mapping `myLong -> java.lang.Long`. So need to check for a value and not the key.

Here's how I would write the whole method:
```java
    private static String mapInt(MappingContext mappingContext, IntValue value, Type<?> graphQLType) {
        if (graphQLType instanceof TypeName) {
            for (Map.Entry<String, String> entry : mappingContext.getCustomTypesMapping().entrySet()) {
                if (isLong(entry.getValue())) {
                    String typeName = ((TypeName) graphQLType).getName();
                    if (isLong(typeName)) {
                        return String.valueOf(value.getValue()).concat("L");
                    }
                }
            }
        }
        if (graphQLType instanceof NonNullType) {
            // unwrapping NonNullType
            return mapInt(mappingContext, value, ((NonNullType) graphQLType).getType());
        }
        return String.valueOf(value.getValue());
    }
```

jxnu-liguobin: > @jxnu-liguobin
> 
> 1. Could you please explain what are you trying to get from `graphQLType fields`?
> 2. In `customTypesMapping` you are checking for a key `Long`, but someone can have scalar called `myLong` and a mapping `myLong -> java.lang.Long`. So need to check for a value and not the key.
> 
> Here's how I would write the whole method:
> 
> ```java
>     private static String mapInt(MappingContext mappingContext, IntValue value, Type<?> graphQLType) {
>         if (graphQLType instanceof TypeName) {
>             for (Map.Entry<String, String> entry : mappingContext.getCustomTypesMapping().entrySet()) {
>                 if (isLong(entry.getValue())) {
>                     String typeName = ((TypeName) graphQLType).getName();
>                     if (isLong(typeName)) {
>                         return String.valueOf(value.getValue()).concat("L");
>                     }
>                 }
>             }
>         }
>         if (graphQLType instanceof NonNullType) {
>             // unwrapping NonNullType
>             return mapInt(mappingContext, value, ((NonNullType) graphQLType).getType());
>         }
>         return String.valueOf(value.getValue());
>     }
> ```

1. oh, it can use forced conversion instead of reflect, thanks.
2. beacuse i only want to solve `Long` in [graphql-java-scalar](https://github.com/graphql-java/graphql-java-extended-scalars), https://github.com/graphql-java/graphql-java-extended-scalars/blob/6f82be9d81b00a580f79efc12b47a0970b12f768/src/main/java/graphql/scalars/ExtendedScalars.java#L213 not customtype like myLong,xxLong.If you think this is better, i can change it.

kobylynskyi: To 2: In case clients are using custom scalar Long, they will need to specify it in the customTypesMapping anyway. Otherwise all unknown scalars are mapped to String by default

PR #297, opened 2020-08-21
Support code generation based on introspection query response #165

### LINKED ISSUE #165: Generate client code from introspection query response
Always same context: we use this project to test our GraphQL API with RestAssured. We use your maven plugin for generating client code from schema file, before compiling and running RestAssured tests.

From our pom.xml for example: 
```xml
          <!-- Generate GraphQL client classes from schema. -->
          <plugin>
            <groupId>io.github.kobylynskyi</groupId>
            <artifactId>graphql-codegen-maven-plugin</artifactId>
            <version>1.8.0</version>
            <executions>
                <execution>
                    <goals>
                        <goal>generate</goal>
                    </goals>
                    <configuration>
                        <graphqlSchemaPaths>${project.basedir}/src/integration-test/resources/schema.graphqls</graphqlSchemaPaths>
                        <outputDir>${project.build.directory}/generated-test-sources/graphql</outputDir>
                        ...
```

It works great! 
But we have to write these schema file(s) that we use to generate this code client.

Can we imagine:
1. the maven plugin does a HTTP request to the GraphQL API sending an introspection query. 
2. the graphql-java-codegen lib is able to generate the client code from this introspection query response (instead of only GQL schema).

Of course, this will be an additionnal feature to the plugin. The plugin must always support the generation from schema file.

I think it will be relevant, specially for generating client code in order to really use the GraphQL API (and not only for testing). Because, server and client aren't probably in the same project. For example, it can be done with Swagger in command line: https://github.com/swagger-api/swagger-codegen#to-generate-a-sample-client-library
 
However, I don't know if it is easily possible, because the result of introspection query has absolutely not the same format of the schema file. 

Before thinking about details, I want to know if you think that it is relevant and possible ?

The first step (retrieving introspection query response from API) maybe can be done by another maven plugin in order to simplify the feature. For example: https://github.com/cjnygard/rest-maven-plugin (I don't take time to test this plugin, but I'm almost sure we can find a maven plugin which makes the job).

Thanks for your time,
Laura

kobylynskyi:
Resolves #165


src/main/java/com/kobylynskyi/graphql/codegen/GraphQLDocumentParser.java
```diff
@@ -52,6 +54,25 @@ static ExtendedDocument getDocument(MappingConfig mappingConfig, List<String> sc
         return extendedDocumentBuilder.build();
     }
 
+    static ExtendedDocument getDocumentFromIntrospectionResult(MappingConfig mappingConfig, String introspectionResult) throws IOException {
+        String introspectionResultContent = Utils.getFileContent(introspectionResult);
+        Map<String, Object> introspectionResultMap = Utils.OBJECT_MAPPER.readValue(introspectionResultContent,
+                new TypeReference<Map<String, Object>>() {
+                });
+        // unwrapping "data" (in case such GraphQL response supplied)
+        if (introspectionResultMap.containsKey("data") && !introspectionResultMap.containsKey("__schema")) {
```
lmartellotto, line 63: Do we need to check that the introspection result doesn't contain "__schema" in order to get the "data"?

I think it is useless, because the introspection query contains either "data" or "__schema".

It means that, if we have an introspection query like that (which is wrong):

```
{
  "data": ...,
  "__shema": ...
}
```

-> we will ignore the "data" object, because `new IntrospectionResultToSchema().createSchemaDefinition(introspectionResultMap)` will only take the "__schema" object.

I think wee need to do these steps, in this order:
1) If the map contains "data", unwrap as you did ;
2) then, check that the map contains the key "__schema" (otherwise throw exception?) ;
3) call `new IntrospectionResultToSchema().createSchemaDefinition(introspectionResultMap)`

lmartellotto:
I have written two (not vital) comments. If you don't want to apply them, never mind (just let me know then I will approve the PR) 😉 

Thanks a lot for your work and your time! I (almost) can't wait to come back from my holidays to use it in my projects! 😝

src/main/java/com/kobylynskyi/graphql/codegen/GraphQLCodegen.java
```diff
@@ -154,7 +169,12 @@ private static void initDefaultValues(MappingConfig mappingConfig) {
         }
     }
 
-    private static void validateConfigs(MappingConfig mappingConfig) {
+    private void validateConfigs(MappingConfig mappingConfig) {
+        if (schemas != null && introspectionResult != null ||
```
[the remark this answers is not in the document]
kobylynskyi: Thanks! Fixed this.

src/main/java/com/kobylynskyi/graphql/codegen/GraphQLDocumentParser.java
kobylynskyi: Thanks, @lmartellotto !
Just so you know, `new IntrospectionResultToSchema().createSchemaDefinition(introspectionResultMap)` performs the check for `"__schema"`.

kobylynskyi:
@lmartellotto I've applied some minor fixes. Please review the PR. Thanks!

lmartellotto (approved):

lmartellotto (approved):

PR #301, opened 2020-08-27
generate `all$` for construct default projection instance automatically 

jxnu-liguobin:


jxnu-liguobin:
> @jxnu-liguobin thanks for working on this!
> Several notes:
> 
> 1. I would prefer to remove config `generateSelectAll` and generate `all$` method always.
> 2. Rename `projectionMaxDepth` to `responseProjectionMaxDepth` (to align with other configs).
> 3. Generate methods `all$()` instead of `selectAll$`.
> 4. Should we use `<=` intead of `<` in this condition: `if (projectionDepthOnFields.getOrDefault("...", 0) <= maxDepth)`?
> 
> P.S.: In the future please refresh your repo so that PR looks cleaner (instead of having 98 commits)
> 
> Thanks!

1. OK, but will change many test cases.
4. change to`<=` that's OK, but it means it include the **current** and **children**  levels, total level is `max+1`

jxnu-liguobin:
see #301 

I tried to merge commit and use upstream , but because my master is not clean, I deleted my repo and configured upstream again, so the repo associated with this PR is invalid,😅

Issue #300, opened 2020-08-25
generate resolver impl and support `selectAll` in projection


jxnu-liguobin:
**Is your feature request related to a problem? Please describe.**
Hi, since the last time we used dynamic automatic call, we think it is better to use automatic generation completely because there may be performance problems with the proxy. Now we consider that we can also entrust the implementation to the framework generation. Of course, I think this may have exceeded the responsibility of this framework.

I have roughly written a template and implemented it by associating `entity resolver returnType`.
In this way all codes will be fully automated,however, I think it may increase the framework coupling.

**Describe the solution you'd like**
ftl:
```
<#if package?has_content>
    package ${package};


</#if>
import com.fasterxml.jackson.core.type.TypeReference;
import com.kobylynskyi.graphql.codegen.model.graphql.GraphQLOperation;
import com.kobylynskyi.graphql.codegen.model.graphql.GraphQLOperationRequest;
import com.kobylynskyi.graphql.codegen.utils.Utils;
import io.github.dreamylost.OkHttp;
import scala.concurrent.Await;
import scala.concurrent.duration.Duration;
import scala.concurrent.ExecutionContext.Implicits.global;
<#list imports as import>
    import ${import}.*;
</#list>

<#if javaDoc?has_content>
/**
<#list javaDoc as javaDocLine>
 * ${javaDocLine}
</#list>
 */
</#if>
<#if generatedInfo.getGeneratedType()?has_content>
@${generatedInfo.getGeneratedType()}(
    value = "com.kobylynskyi.graphql.codegen.GraphQLCodegen",
    date = "${generatedInfo.getDateTime()}"
)
</#if>
public class ${className}<#if implements?has_content> extends <#list implements as interface>${interface}<#if interface_has_next>, </#if></#list></#if> {

    public ${className}() {
    }

<#list operations as operation>
<#if operation.javaDoc?has_content>
    /**
    <#list operation.javaDoc as javaDocLine>
     * ${javaDocLine}
    </#list>
     */
</#if>
<#if operation.deprecated>
    @Deprecated
</#if>
<#list operation.annotations as annotation>
    @${annotation}
</#list>
    ${operation.type} ${operation.name}(${requestType} operateRequest) throws Exception {
    <#if projectionType?has_content>
        ${projectionType} projection = new ${projectionType}();
        for (Method m : projection.getClass().getDeclaredMethods()) {
            Utils.selectOnProjection(projection, m, 1, ${max});
        }
[...]

PR #302, opened 2020-08-28
Add `all` on projection #300

jxnu-liguobin:


docs/client-side-cases-by-proxy.md
```diff
@@ -9,11 +9,12 @@ However, since we have simplified the client-side logic as much as possible, thi
 so that the users do not need to choose the fields on projection. 
 This will also have a bad effect, we can only use all fields of the projection on the returned structure by default. 
 Moreover, we also need set to default the depth of nested queries so that the proxy can end smoothly. 
-All in all, the scenarios used are limited. Currently `.. on` and `alias` are not supported.
+All in all, the scenarios used are limited. Currently, `.. on` and `alias` not be supported.
```
joffrey-bion, line 12: Why this change? *"X and Y are not supported"* was grammatically correct, *"X and Y not be supported"* is not correct.
I agree with the added comma, but the verb needs to stay conjugated.

jxnu-liguobin: I want to say that this is passive

joffrey-bion: But this *is* the passive voice. The active voice would be *"we don't support X"*. The passive voice is *"X is not supported"*. Using "be" here is just incorrect.

jxnu-liguobin: ok,thanks.

src/main/java/com/kobylynskyi/graphql/codegen/model/graphql/GraphQLResponseProjection.java
```diff
@@ -12,6 +10,36 @@
...
+     * }}
+     * Map Notes:
+     * `key` is parentProjection.childProjection.currentMethod. e.g. `CharacterResponseProjection.CharacterResponseProjection.friends` (excluding the first layer, so if only want the first child layer, use `all$(1)`)
+     * `value` is current depth for Character type. Each projection has a new instance of `projectionDepthOnFields`, so it always be `1` or `0`.
+     * and `responseProjectionMaxDepth` will reduce by recursive.
+     */
+    protected final Map<String, Integer> projectionDepthOnFields = new HashMap<>();
+
+    //Defined at the parent level to use dynamic calls, default null.
+    public GraphQLResponseProjection all$() {
```
kobylynskyi, line 35: I think we should use:
`public abstract GraphQLResponseProjection all$();`
`public abstract GraphQLResponseProjection all$(int maxDepth);`

kobylynskyi (changes requested):

src/main/java/com/kobylynskyi/graphql/codegen/model/graphql/GraphQLResponseProjection.java
jxnu-liguobin: OK😀

kobylynskyi (approved):

PR #307, opened 2020-09-04
Update gradle plugin readme

blocha:
In my own project we migrated from Groovie to Kotlin gradle files, and, sadly, info from readme file did not work. This is the way we can add graphql generating task to java compile in the project when we use kotlin in gradle files. 

Cheers


kobylynskyi (approved):
@blocha thank you so much for spotting the issue and fixing it!

PR #350, opened 2020-10-09
Support multiple queries in a single request #340

### LINKED ISSUE #340: Combining Multiple Queries in Request
GraphQL supports multiple queries to be run in one call. In this case since GraphQL Request only accepts one input it is difficult to pass multiple request and projections especially in case of RestTemplate Multiple GraphQL Query will also give multiple combined response Type
 
A wrapper Class on GraphQLRequest and GraphQLResponse to combine multiple request and handle call and response

kobylynskyi:
@joffrey-bion, @lmartellotto, @logicatmidod, could you please review? Thanks!

---
Closes #340

### DIFF
--- a/plugins/gradle/example-client/build.gradle
+++ b/plugins/gradle/example-client/build.gradle
@@ -92,7 +92,7 @@ task graphqlCodegenStarwarsService(type: GraphQLCodegenGradleTask) {
     ]
     generateClient = true
     generateApis = false
-    generateBuilder = false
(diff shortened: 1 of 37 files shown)


logicatmidod (approved):
This will work I have tried at my end by replicating the code. But need to check if some how response wrapper also can be created for managing these responses. Though we should be providing ability to register Converters to those responses. That can convert to the appropriate Objects.

PR #402, opened 2020-11-19
Scala

jxnu-liguobin:
1. Use the scala collection `Seq` whenever possible
2. Use Scala's basic types as much as possible, and use Java's object type only when data can be null
3. Immutable data structure is used by default
4. Keep builder mode
5. Unlike Java, enum must be imported into themselves
6. All types use the case class
7. Use Scala `Option` instead of Java `Optional`
8. Also add in maven and gradle plugins
9. Use `@BeanProperty` to override java get/set methods
10. The default indentation uses four spaces, which is different from most Scala standard code, and may be changed later or introduce automatic formatting to avoid precise template configuration

and ...

In addition to the unit test, this code has been verified by all schema I have used, and generated 1000 Scala files.
The next step will provide the sample code used in the SBT plugin.

but only a draft


kobylynskyi (dismissed):
@jxnu-liguobin, that's great!
You did an awesome job.

kobylynskyi:
@jxnu-liguobin, I've introduced some new changes recetly. Will take care of resolving conflicts...

PR #411, opened 2020-12-03
Refactor scala tostring

jxnu-liguobin:
This update is as follows

1. Refactoring `toString` and `equals` methods(Simplify code bloat caused by large number of fields, decrease by 50%).
2. For Scala(after that, kotlin will do the same), the `builder` is no longer generated by default.
3. Fix bug when set`generatebuilder:=false`, the companion object was not generated.
4. Template support `serializeUsingObjectMapper`
5. remove invalid `if` in `scalaClassGraphqlInterface.ftl` and `scalaClassGraphqlEnum.ftl`

Please check to see if `toString` keeps its original function(ParametrizedInput and Type). such as
src/test/resources/expected-classes/scala/tostring/QueryPrivateParametrizedInput.scala.txt
src/test/resources/expected-classes/scala/Commit.scala.txt


kobylynskyi (approved):
@jxnu-liguobin, looks good! Will release in 4.0.1 this week

PR #417, opened 2020-12-07
Scala optimization

jxnu-liguobin:
1.Remove the scala `this` constructor and use `apply` in request class
2.Change ParametrizedInput to case class, so `equals` and `hashcode` can be deleted
3.Delete invalid Java classes import
4.Define the interface method as a property, and do not use `@beanproperty` for Java get/set, Compatibility for compatibility will lead to ugly Scala code.
Case class always generates `toString`、`equals`、`hashcode` by default. I don't think this is always a good way for non data classes. So ,At present, `request, response, response projection` do not need to be designed as case class.

now, immutable class
```scala
trait TunnelConfig {
  val s:Int
}

class A (val s:Int) extends TunnelConfig
```
mutable class
```scala
trait TunnelConfig {
  var s:Int
}

class A (var s:Int) extends TunnelConfig
```


kobylynskyi (approved):
@jxnu-liguobin, do you please to introduce more Scala-related changes soon?
Please let me know when I can release graphql-java-codegen-4.0.1.

jxnu-liguobin:
> @jxnu-liguobin, do you please to introduce more Scala-related changes soon?
> Please let me know when I can release graphql-java-codegen-4.0.1.

Template level optimization has been completed. There are two known minor problems need to wait for the refactoring.
1.The default value of the plugin causes initdefaultValues doesnot set ImmutableModels=true by default in scala
2.ImmutableModels  donot need `val` on properties

jxnu-liguobin:
Major changes

1. Refactoring `toString` and `equals` methods.
2. Use **case class** for `XXParametrizedInput` and create apply for XXGraphQLOperationRequest
3. Change the field of graphql interface from Scala method to scala property.
4. Fixed wrong request class generated when not building builder.

Issue #407, opened 2020-11-27
Refactor library for multi-language support


jxnu-liguobin:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

When I implemented Scala, I found that the existing code is not very convenient to implement multi language, resulting in a lot of judgment logic all over the mapper. Shall we consider refactoring it? (different implementations for different languages, instead of using `if else` in only one mapper). 

**Describe alternatives you've considered**

**Additional context**

kobylynskyi:
@jxnu-liguobin, do you plan to introduce more changes in 4.0.1, or I can release it?

jxnu-liguobin:
> @jxnu-liguobin, do you plan to introduce more changes in 4.0.1, or I can release it?

At present,there are no urgent problems to be solved. I think it can be released. Next time, kotlin should be at least 4.1.0 or 5.0.0.

PR #418, opened 2020-12-08
Refactor for multi-language support #407

kobylynskyi:
@jxnu-liguobin, please take a look.
Now all language-specific code is residing in its own classes:
image


src/main/java/com/kobylynskyi/graphql/codegen/scala/ScalaGraphQLTypeMapper.java
```diff
@@ -0,0 +1,127 @@
...
+     * {@inheritDoc}
+     */
+    @Override
+    public String mapDirectiveArgumentValue(MappingContext mappingContext, Argument dirArg, String argumentValueFormatter) {
+        return valueMapper.map(mappingContext, dirArg.getValue(), null, argumentValueFormatter);
+    }
+
+    @Override
+    public boolean addModelValidationAnnotationForType(String possiblyPrimitiveType) {
+        return !JavaGraphQLTypeMapper.isJavaPrimitive(possiblyPrimitiveType); // reconsider
```
jxnu-liguobin, line 124: isPrimitive ?

src/main/java/com/kobylynskyi/graphql/codegen/scala/ScalaGraphQLCodegen.java
```diff
@@ -0,0 +1,81 @@
...
+    protected void initDefaultValues(MappingConfig mappingConfig) {
+        if (mappingConfig.getGenerateBuilder() == null) {
+            // functional expression
+            mappingConfig.setGenerateBuilder(false);
+        }
+        if (mappingConfig.getGenerateImmutableModels() == null) {
+            // functional expression
+            mappingConfig.setGenerateImmutableModels(true);
+        }
+        super.initDefaultValues(mappingConfig);
```
jxnu-liguobin, line 62: `super.initDefaultValues(mappingConfig);`  Maybe it should be at the top?

jxnu-liguobin:
It looks good. I've changed the SBT plugin

src/main/java/com/kobylynskyi/graphql/codegen/scala/ScalaGraphQLTypeMapper.java
kobylynskyi: Yes! That's what was present in the original non-refactored code.
Scala tests were failing because of this.

src/main/java/com/kobylynskyi/graphql/codegen/scala/ScalaGraphQLCodegen.java
kobylynskyi: If I move it to the top then `super.initDefaultValues(mappingConfig)` will set values for `setGenerateImmutableModels` and `setGenerateBuilder`, but as I understand, you don't want that.

jxnu-liguobin: Yes, I understand that will override the settings

src/main/java/com/kobylynskyi/graphql/codegen/scala/ScalaGraphQLTypeMapper.java
jxnu-liguobin: No problem at the moment, I'll optimize it later.(Annotations on Scala primitive types should be removed from the test and code.)

jxnu-liguobin (approved):

Issue #420, opened 2020-12-08
[Scala] Optimization in the  future


jxnu-liguobin:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

- Scala code does not use Java types at all, such as nullable Int, will change it from `Integer` =>  `Option[Int]`
- Removing the extra declare `val` on the case class when generate imutable classes.This is the same as kotlin. It needs to add the context of relevant class files to filter or judge, **but it is optional in scala and necessary in kotlin.**
- nonull annotation on Scala primitive types should be removed from the code.

**Describe alternatives you've considered**

**Additional context**

Issue #15, opened 2019-11-24
Kotlin support


mpg57:
Are there any plans for this tool to support Kotlin class generation?

kobylynskyi:
Hey @mpg57, no plans yet. Sorry.

joffrey-bion:
I'd also love to have the option to generate Kotlin classes from that.
If I have time at some point, I might be able to contribute the relevant template files.

jxnu-liguobin:
> I'd also love to have the option to generate Kotlin classes from schemas.
> If I have time at some point, I might be able to contribute the relevant template files.

Hi,  If  can provide templates for other jvm language, can be used them directly without writing any other code?

jxnu-liguobin:
kotlin templates have done(draft): https://github.com/jxnu-liguobin/graphql-java-codegen/blob/kotlin/src/main/resources/templates/kotlin-lang

If feasible, after the refactoring is completed, this template will be used as an implementation of Kotlin

**May not be the best implementation. Because I have no actual kotlin development experience.**
I made reference to Scala.

Known issues:  https://github.com/kobylynskyi/graphql-java-codegen/issues/415
Known problem: subclass  muat have `override` if superclass have abstract method.
There is no solution for the time being, prepare to record all the methods of the graphql interface, and judge in the subclass whether you need to `override`.

There are some examples.
type 1
```kotlin
package com.github.graphql

import com.kobylynskyi.graphql.codegen.model.graphql.GraphQLRequestSerializer
import java.util.StringJoiner

@javax.annotation.Generated(
    value = ["com.kobylynskyi.graphql.codegen.GraphQLCodegen"],
    date = "2020-12-31T23:59:59-0500"
)
open class AcceptTopicSuggestionInput() {

    private lateinit var clientMutationId: String
    @javax.validation.constraints.NotNull
    private lateinit var name: String
    @javax.validation.constraints.NotNull
    private lateinit var repositoryId: String

    constructor(clientMutationId: String, name: String, repositoryId: String): this() {
        this.clientMutationId = clientMutationId
        this.name = name
        this.repositoryId = repositoryId
    }

    companion object {
        fun builder(): Builder = Builder()
    }

    fun getClientMutationId(): String = clientMutationId

    fun getName(): String = name

    fun getRepositoryId(): String = repositoryId

    override fun equals(other: Any?): Boolean {
        if (this === other) {
            return true
        }
        if (other == null || javaClass != other.javaClass) {
            return false
[...]

PR #426, opened 2020-12-09
Kotlin support #15

jxnu-liguobin:
This is a preview of the draft pr.
1.unit test is not enough
2.have not write examples for used
3.have not verification is conducted
4.It fully supports kotlin's nullable type,there will be no such problem.https://github.com/kobylynskyi/graphql-java-codegen/issues/415,Scala will do the same in the future.

Using `mandatory`, Scala / kotlin can be better modeled without any annotations. At the same time, it supports option type naturally, And make the semantics of graphql exactly the same as the generated code

for Scala 
1.Remove invalid import enumeration imports
2.Remove annotations on Scala primitive types and fix unit test
3.Keyword conflicts are resolved using ”``“, while conflicts with methods are still consistent
4.Due to the use of `case class` - scala / `data class` - kotlin, the `toString` of ParameterizedInput has been removed
5.Add Scala's empty type unit test
6.Scala optimization, we no longer need to rely on the creation order of files.
7.For primitive type canbe null, use `scala.Option` https://github.com/kobylynskyi/graphql-java-codegen/issues/420

I created about a thousand files using the production schema,Although it looks good and compiles, unit test is still not enough.

```
[info] Compiling 1035 Scala sources to /Users/liguobin/project/growingio-graphql-javasdk/target/scala-2.12/classes ...
[success] Total time: 48 s, completed 2020-12-11 19:38:47
```


src/main/java/com/kobylynskyi/graphql/codegen/kotlin/KotlinDataModelMapper.java
```diff
@@ -0,0 +1,61 @@
...
+    private static final Set<String> KOTLIN_RESTRICTED_METHOD_NAMES = new HashSet<>(Arrays.asList("notify", "notifyAll", "wait"));
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String capitalizeIfRestricted(MappingContext mappingContext, String fieldName) {
+
+        if (KOTLIN_RESTRICTED_KEYWORDS.contains(fieldName)) {
+            return Utils.capitalize(fieldName);
```
joffrey-bion, line 33: This is unnecessary, Kotlin supports backquoting names if you need to define identifiers with special characters or that are reserved words. I think we should use that instead.

src/main/java/com/kobylynskyi/graphql/codegen/kotlin/KotlinDataModelMapper.java
```diff
@@ -0,0 +1,61 @@
...
+        return fieldName;
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public String capitalizeMethodNameIfRestricted(MappingContext mappingContext, String methodName) {
+        if (KOTLIN_RESTRICTED_KEYWORDS.contains(methodName)) {
+            return Utils.capitalize(methodName);
```
joffrey-bion, line 44: Same thing here for method names equal to keywords, I think we should prefer backquotes here.

src/main/java/com/kobylynskyi/graphql/codegen/kotlin/KotlinDataModelMapper.java
jxnu-liguobin: yes,

src/main/java/com/kobylynskyi/graphql/codegen/kotlin/KotlinDataModelMapper.java
```diff
@@ -0,0 +1,61 @@
...
+        }
+        if (KOTLIN_RESTRICTED_METHOD_NAMES.contains(methodName)) {
+            return Utils.capitalize(methodName);
```
joffrey-bion, line 47: **_ATTENTION_**: in this case we can't use backquotes, I think platform method conflicts cannot be solved with backquotes, so we do need to keep the capitalization for this case.

src/main/java/com/kobylynskyi/graphql/codegen/kotlin/KotlinGraphQLTypeMapper.java
```diff
@@ -0,0 +1,151 @@
...
+     * @return String The name of the the wrapped type.
+     */
+    @Override
+    public String wrapSuperTypeIntoList(MappingContext mappingContext, String type) {
+        return getGenericsString(mappingContext, KOTLIN_UTIL_LIST, "out " + type);
+    }
+
+    /**
+     * {@inheritDoc}
+     */
```
joffrey-bion, line 68: These doc comments are unnecessary. Javadoc is inherited by default. We only need `@inheritDoc` if we need to write other javadoc but still include the parent's doc as far as I remember.

src/main/java/com/kobylynskyi/graphql/codegen/kotlin/KotlinGraphQLTypeMapper.java
```diff
@@ -0,0 +1,151 @@
...
+    @Override
+    public String mapDirectiveArgumentValue(MappingContext mappingContext, Argument dirArg, String argumentValueFormatter) {
+        return valueMapper.map(mappingContext, dirArg.getValue(), null, argumentValueFormatter);
+    }
+
+    @Override
+    public boolean addModelValidationAnnotationForType(String possiblyPrimitiveType) {
+        return !KotlinGraphQLTypeMapper.isKotlinPrimitive(possiblyPrimitiveType);
```
joffrey-bion, line 125: Kotlin doesn't have primitive types. I'm not sure this is valid.
I don't remember if this library supports other validations than NotNull.
If we do support arbitrary validation, it should be done on any type (including 'primitive').
If we only support nullability, it is irrelevant in Kotlin because we have the type system for that already.

src/main/java/com/kobylynskyi/graphql/codegen/kotlin/KotlinGraphQLTypeMapper.java
```diff
@@ -0,0 +1,151 @@
...
+    public static String defaultValueKotlinPrimitive(String kotlinType) {
+        switch (kotlinType) {
+            case "Long":
+                return "0L";
+            case "Float":
+                return "0F";
+            case "Double":
+                return "0D";
+            case "Char":
+                return "''";
```
joffrey-bion, line 137: This is not a valid `Char`. Default could be `0.toChar()` maybe?

src/main/java/com/kobylynskyi/graphql/codegen/mapper/RequestResponseDefinitionToDataModelMapper.java
```diff
@@ -154,13 +157,19 @@ private static ProjectionParameterDefinition getTypeNameProjectionParameterDefin
         String className = DataModelMapper.getParametrizedInputClassName(mappingContext, fieldDefinition, parentTypeDefinition);
         Map<String, Object> dataModel = new HashMap<>();
         // ParametrizedInput classes are sharing the package with the model classes, so no imports are needed
+        List<ParameterDefinition> fields = inputValueDefinitionToParameterMapper.map(
+                mappingContext, fieldDefinition.getInputValueDefinitions(), parentTypeDefinition.getName());
+        //kotlin not support empty
+        if (mappingContext.getGeneratedLanguage().equals(GeneratedLanguage.KOTLIN) &&
```
joffrey-bion, line 163: Kotlin does support classes with no fields, we just need to use regular classes (not data classes) for them.
This is because equals, hashcode etc don't make sense if there are no fields.

src/main/java/com/kobylynskyi/graphql/codegen/mapper/TypeDefinitionToDataModelMapper.java
```diff
@@ -79,20 +69,27 @@ private static ParameterDefinition merge(ParameterDefinition typeDef, ParameterD
         ExtendedDocument document = mappingContext.getDocument();
 
         Map<String, Object> dataModel = new HashMap<>();
+        Collection<ParameterDefinition> fields = getFields(mappingContext, definition, document);
+        //kotlin not support empty
+        if (mappingContext.getGeneratedLanguage().equals(GeneratedLanguage.KOTLIN) &&
+                fields.isEmpty()) {
+            throw new UnableToCreateFileException(new Exception("Type requires at least one field to use in Kotlin."));
+        }
```
joffrey-bion, line 77: Again, we just need to use a regular `class` or `object` in this case in Kotlin, just not a `data class`.

src/main/java/com/kobylynskyi/graphql/codegen/kotlin/KotlinGraphQLTypeMapper.java
jxnu-liguobin: Just to not verify null.if it's a `a:Int`, then, donot need generate if (a!=null)
Of course, the question mark is also judged. It can be deleted here, but in order to allow the annotation to be verified, it is still retained.

src/main/resources/templates/kotlin-lang/kotlinClassGraphqlEnum.ftl
```diff
@@ -0,0 +1,40 @@
...
+</#if>
+<#if field.deprecated>
+    @java.lang.Deprecated
+</#if>
+    ${field.javaName}("${field.graphqlName}")<#if field_has_next>,
+<#else></#if>
+</#list>
+</#if>
+
+}
```
joffrey-bion, line 40: This is just my OCD speaking, but shouldn't we terminate the file with line feed?

src/main/resources/templates/kotlin-lang/kotlinClassGraphqlParametrizedInput.ftl
```diff
@@ -0,0 +1,59 @@
...
+        val joiner = StringJoiner(", ", "(", ")")
+<#if fields?has_content>
+<#list fields as field>
+        <#if field.type?ends_with("?")>
+            <#if MapperUtil.isKotlinPrimitive(field.type)>
+        joiner.add("${field.originalName}: " + GraphQLRequestSerializer.getEntry(${field.name}))
+            <#else>
+        if (${field.name} != null) {
+            joiner.add("${field.originalName}: " + GraphQLRequestSerializer.getEntry(${field.name}))
+        }
```
joffrey-bion, line 51: Could you please explain this part? Why would the field name be null when the field is not a primitive type?

src/main/resources/templates/kotlin-lang/kotlinClassGraphqlRequest.ftl
```diff
@@ -0,0 +1,149 @@
...
+        <#list fields as field>
+        <#if field.defaultValue?has_content>
+        private var ${field.name}: ${field.type} = ${field.defaultValue}
+        <#else>
+        <#if field.type?ends_with("?")>
+        private var ${field.name}: ${field.type} = null
+        <#else>
+        <#if MapperUtil.isKotlinPrimitive(field.type)>
+        <#assign default = MapperUtil.defaultValueKotlinPrimitive(field.type)/>
+        private var ${field.name}: ${field.type} = default
```
joffrey-bion, line 104: If the field is mandatory and doesn't have a default value, why would we arbitrarily add a default just because the type happens to be a primitive? I don't think we should do that

src/main/resources/templates/kotlin-lang/kotlinClassGraphqlRequest.ftl
```diff
@@ -0,0 +1,149 @@
...
+        <#list fields as field>
+            <#if field.javaDoc?has_content>
+        /**
+        <#list field.javaDoc as javaDocLine>
+         * ${javaDocLine}
+        </#list>
+         */
+    </#if>
+    <#if field.deprecated>
+        @java.lang.Deprecated
```
joffrey-bion, line 128: I think we should use the Kotlin equivalent here.

src/test/resources/expected-classes/kt/AddLabelsToLabelableInput.kt.txt
```diff
@@ -0,0 +1,28 @@
...
+        val joiner = StringJoiner(", ", "{ ", " }")
+        if (clientMutationId != null) {
+            joiner.add("clientMutationId: " + GraphQLRequestSerializer.getEntry(clientMutationId))
+        }
+        joiner.add("labelIds: " + GraphQLRequestSerializer.getEntry(labelIds))
+        if (labelableId != null) {
+            joiner.add("labelableId: " + GraphQLRequestSerializer.getEntry(labelableId))
+        }
+        return joiner.toString()
+    }
```
joffrey-bion, line 27: Why do we generate a `toString()` method here? Kotlin's `data class` already has one.

src/test/resources/expected-classes/kt/AddLabelsToLabelableMutationResolver.kt.txt
```diff
@@ -0,0 +1,13 @@
+package com.github.graphql
+
+
+@javax.annotation.Generated(
+    value = ["com.kobylynskyi.graphql.codegen.GraphQLCodegen"],
+    date = "2020-12-31T23:59:59-0500"
+)
+interface AddLabelsToLabelableMutationResolver {
+
+    @Throws(Exception::class)
```
joffrey-bion, line 10: Do we want this?

src/test/resources/expected-classes/kt/Commit_Var_Field.kt.txt
```diff
@@ -0,0 +1,114 @@
...
+        joiner.add("tarballUrl: " + GraphQLRequestSerializer.getEntry(tarballUrl))
+        joiner.add("tree: " + GraphQLRequestSerializer.getEntry(tree))
+        joiner.add("treeResourcePath: " + GraphQLRequestSerializer.getEntry(treeResourcePath))
+        joiner.add("treeUrl: " + GraphQLRequestSerializer.getEntry(treeUrl))
+        joiner.add("url: " + GraphQLRequestSerializer.getEntry(url))
+        joiner.add("viewerCanSubscribe: " + GraphQLRequestSerializer.getEntry(viewerCanSubscribe))
+        joiner.add("viewerSubscription: " + GraphQLRequestSerializer.getEntry(viewerSubscription))
+        joiner.add("zipballUrl: " + GraphQLRequestSerializer.getEntry(zipballUrl))
+        return joiner.toString()
+    }
```
joffrey-bion, line 113: Again ,I don't believe we need these toString method overrides.

src/main/resources/templates/kotlin-lang/kotlinClassGraphqlParametrizedInput.ftl
jxnu-liguobin: This is the same as Java

src/main/resources/templates/kotlin-lang/kotlinClassGraphqlRequest.ftl
jxnu-liguobin: not find a good way to implement builder set..

src/main/resources/templates/kotlin-lang/kotlinClassGraphqlRequest.ftl
jxnu-liguobin: It's not going to be used for the time being, because kotlin needs some additional parameters

src/main/java/com/kobylynskyi/graphql/codegen/kotlin/KotlinGraphQLTypeMapper.java
joffrey-bion: I have trouble understanding what you're saying here. But yeah as far as nullability is concerned, there is no difference between primitive types and non-primitive types in Kotlin. Only `?` matters.

What annotation are you talking about? `javax.validation.NotNull`?
This one is also not needed in the Kotlin codegen.

src/test/resources/expected-classes/kt/AddLabelsToLabelableInput.kt.txt
jxnu-liguobin: This is used to process graphql query and must be rewritten

src/test/resources/expected-classes/kt/AddLabelsToLabelableMutationResolver.kt.txt
jxnu-liguobin: Just for the same function as Java and scala

src/main/resources/templates/kotlin-lang/kotlinClassGraphqlParametrizedInput.ftl
joffrey-bion: Then this is wrong I believe. It makes sense in Java because primitive = non null, and non-primitive = nullable. The check to decide whether to put a null check or not in Kotlin should be entirely determine by whether the type is a nullable type or not (`?`) and should not have anything to do with primitives.

src/test/resources/expected-classes/kt/AddLabelsToLabelableMutationResolver.kt.txt
joffrey-bion: This annotation only affects Java callers of the Kotlin code. So I'm not sure it matters much here.

src/main/resources/templates/kotlin-lang/kotlinClassGraphqlRequest.ftl
joffrey-bion: I think all you need is a message here. So we could have a default message like "this is deprecated in GraphQL".

But if I remember correctly, the reason for deprecation is also mandatory in the GraphQL schema, right?
Then we should use that same reason message here in Kotlin.

src/main/resources/templates/kotlin-lang/kotlinClassGraphqlRequest.ftl
joffrey-bion: Isn't the last `else` part dealing with this already for any other type?

src/test/resources/expected-classes/kt/AddLabelsToLabelableInput.kt.txt
joffrey-bion: Ah I see, so the actual string has meaning, and is not "just a `toString`", right?
In that case I understand.
I guess we should rename the method instead of using `toString` for business logic.

src/main/java/com/kobylynskyi/graphql/codegen/kotlin/KotlinGraphQLTypeMapper.java
jxnu-liguobin: There is no disable nonull annotation, it is only turned off by default. Of course, if it is completely prohibited, it can be removed

src/main/resources/templates/kotlin-lang/kotlinClassGraphqlParametrizedInput.ftl
jxnu-liguobin: In fact, I'm not sure if this method is used for the request api, otherwise I think we should remove the method.

src/main/resources/templates/kotlin-lang/kotlinClassGraphqlRequest.ftl
jxnu-liguobin: lateinit cannot use in primitive types ，so set a default value for it. And  isKotlinPrimitive will be removed

src/main/resources/templates/kotlin-lang/kotlinClassGraphqlRequest.ftl
jxnu-liguobin: At present, it seems that it is not necessary, but I will try the default message

src/test/resources/expected-classes/kt/AddLabelsToLabelableInput.kt.txt
jxnu-liguobin: This may need to be modified greatly. It is suggested that it should not be changed at first

src/test/resources/expected-classes/kt/AddLabelsToLabelableMutationResolver.kt.txt
jxnu-liguobin: There are cross language possibilities. Just like us in Scala,.we only focus on Scala, but have Java users. So consider generating Java or Scala both need another language.

src/main/resources/templates/kotlin-lang/kotlinClassGraphqlRequest.ftl
joffrey-bion: Then my apologies, I really didn't know that! So yeah the JVM details seem to leak into the language in some places. My bad.

src/main/resources/templates/kotlin-lang/kotlinClassGraphqlRequest.ftl
joffrey-bion: It looks like the reason is an optional parameter in the `@deprecated` directive in GraphQL, but there is a default reason in that case, which is "No longer supported". So I think there is always a reason that we could use here.

src/main/java/com/kobylynskyi/graphql/codegen/GraphQLCodegen.java
```diff
@@ -255,9 +255,13 @@ private void validateConfigs(MappingConfig mappingConfig) {
         MappingContext context = new MappingContext(mappingConfig, document, generatedInformation);
 
         List<File> generatedFiles = new ArrayList<>();
+        // Enumeration must be first for scala, interface must before type for kotlin/scala!!!
```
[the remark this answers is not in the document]
jxnu-liguobin: Because Scala's enumerations need to be imported for use, they may not be generated before the type is generated

kotlin also, for override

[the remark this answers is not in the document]
jxnu-liguobin: You need to know which interfaces need to be rewritten to add 'override'. Otherwise, it cannot be implemented.
But Scala and Java's `override` is optional, which is also due to kotlin's strict restrictions

src/main/java/com/kobylynskyi/graphql/codegen/mapper/GraphQLTypeMapper.java
```diff
@@ -1,5 +1,6 @@
 package com.kobylynskyi.graphql.codegen.mapper;
 
+import com.kobylynskyi.graphql.codegen.model.GeneratedLanguage;
```
kobylynskyi, line 3: @jxnu-liguobin, please remove unused import

src/main/java/com/kobylynskyi/graphql/codegen/mapper/RequestResponseDefinitionToDataModelMapper.java
```diff
@@ -196,6 +197,12 @@ private static ProjectionParameterDefinition getTypeNameProjectionParameterDefin
         dataModel.put(METHOD_NAME, dataModelMapper.capitalizeMethodNameIfRestricted(mappingContext, operationDef.getName()));
         dataModel.put(RETURN_TYPE_NAME, javaType);
         dataModel.put(GENERATED_INFO, mappingContext.getGeneratedInformation());
+        // if getUseOptionalForNullableReturnTypes == true, append `?` for kotlin
```
kobylynskyi, line 200: @jxnu-liguobin, I think it is better if you override `getLanguageType` inside `KotlinGraphQLTypeMapper` and append "?" there:

```java
    @Override
    public NamedDefinition getLanguageType(MappingContext mappingContext, String graphQLType, String name, String parentTypeName, boolean mandatory, boolean collection) {
        NamedDefinition languageType = GraphQLTypeMapper.super.getLanguageType(mappingContext, graphQLType, name, parentTypeName, mandatory, collection);
        if (mappingContext.getUseOptionalForNullableReturnTypes().equals(Boolean.TRUE)) {
            languageType.setJavaName(languageType.getJavaName() + "?");
        }
        return languageType;
    }
```

Then this will take affect in all underlying Mappers.

src/main/java/com/kobylynskyi/graphql/codegen/utils/Utils.java
```diff
@@ -231,4 +231,13 @@ public static String substringBetween(final String str, final String open, final
         return null;
     }
 
+    /**
+     * for scala/kotlin
+     * @param fieldName field name in RESTRICTED_KEYWORDS  or RESTRICTED_METHOD_NAMES
+     * @return new field name
+     */
+    public static String wrapperFieldName(String fieldName){
```
kobylynskyi, line 239: Please make it more generic as I might reuse it in other places:
```java
    public static String wrapString(String str, String wrapWith) {
        if (str == null || wrapWith == null) {
            return str;
        }
        return wrapWith + str + wrapWith;
    }
```

kobylynskyi (changes requested):

src/main/java/com/kobylynskyi/graphql/codegen/mapper/RequestResponseDefinitionToDataModelMapper.java
jxnu-liguobin: yes,

jxnu-liguobin: I will extract a`getResponseLanguageType` and rewrite it in different languages, What's your opinion?

src/main/java/com/kobylynskyi/graphql/codegen/mapper/InterfaceDefinitionToDataModelMapper.java
```diff
@@ -44,15 +40,21 @@ public InterfaceDefinitionToDataModelMapper(GraphQLTypeMapper graphQLTypeMapper,
...
+        dataModel.put(CLASS_NAME, clazzName);
         dataModel.put(JAVA_DOC, definition.getJavaDoc());
         dataModel.put(ANNOTATIONS, graphQLTypeMapper.getAnnotations(mappingContext, definition));
-        dataModel.put(FIELDS, fieldDefinitionToParameterMapper.mapFields(
-                mappingContext, definition.getFieldDefinitions(), definition));
+        dataModel.put(FIELDS, fields);
         dataModel.put(GENERATED_INFO, mappingContext.getGeneratedInformation());
         dataModel.put(ENUM_IMPORT_IT_SELF_IN_SCALA, mappingContext.getEnumImportItSelfInScala());
         dataModel.put(IMMUTABLE_MODELS, mappingContext.getGenerateImmutableModels());
+        if (mappingContext.getParentInterfacePropertiesInKotlin().containsKey(clazzName)) {
```
kobylynskyi, line 53: This is the only thing I have doubts about. Any ideas on how to improve it so that we can execute mappers independently?
Maybe (just for Kotlin) we can precompute these values beforehand? (In the similar way how I did in `ExtendedDocument.getTypesUnionsInterfacesNames()`)

kobylynskyi (changes requested):

src/main/java/com/kobylynskyi/graphql/codegen/mapper/InterfaceDefinitionToDataModelMapper.java
jxnu-liguobin: Does the document contain all the content? If so, it can be obtained directly from the document, and scala's `enumimportitself` in scala can do the same

src/main/java/com/kobylynskyi/graphql/codegen/GraphQLCodegen.java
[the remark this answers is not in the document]
jxnu-liguobin: Use document to pre calculate and no longer rely on order.

src/main/java/com/kobylynskyi/graphql/codegen/model/MappingContext.java
```diff
@@ -275,7 +280,44 @@ public GeneratedInformation getGeneratedInformation() {
...
 
     public Set<String> getEnumImportItSelfInScala() {
+        // Only for scala
+        if (GeneratedLanguage.SCALA.equals(this.config.getGeneratedLanguage()) && enumImportItSelfInScala == null) {
+            enumImportItSelfInScala = this.document.getEnumDefinitions().parallelStream().map(this::getModelClassNameWithPrefixAndSuffix).collect(Collectors.toSet());
+        }
         return enumImportItSelfInScala;
     }
 
+    public Map<String, Set<String>> getParentInterfaceProperties() {
```
kobylynskyi, line 290: I might refactor this in the future to have a separate `MappingContext`s for diff languages.
But for now looks ok!

kobylynskyi (approved):

src/main/java/com/kobylynskyi/graphql/codegen/model/MappingContext.java
jxnu-liguobin: Today I made a improvement for kotlin / Scala (but not implement in Scala)

kobylynskyi (dismissed):

kobylynskyi:
@jxnu-liguobin, looks good to me. 👍 
@joffrey-bion, could you please also cross-check this PR? Thank you!

src/main/java/com/kobylynskyi/graphql/codegen/kotlin/KotlinGraphQLTypeMapper.java
```diff
@@ -0,0 +1,145 @@
...
+                                              String parentTypeName) {
+        String computedTypeName = namedDefinition.getJavaName();
+        if (parentTypeName.equalsIgnoreCase(GraphQLOperation.SUBSCRIPTION.name()) &&
+                Utils.isNotBlank(mappingContext.getSubscriptionReturnType())) {
+            // in case it is subscription and subscriptionReturnType is set
+            return getGenericsString(mappingContext, mappingContext.getSubscriptionReturnType(), computedTypeName);
+        }
+        // Consider only Java-kotlin OR Java-Scala cross language calls
+        if (Boolean.TRUE.equals(mappingContext.getUseOptionalForNullableReturnTypes()) && !namedDefinition.isMandatory()) {
+            if (!computedTypeName.startsWith(KOTLIN_UTIL_LIST) && !computedTypeName.startsWith(JAVA_UTIL_LIST)) {
```
joffrey-bion, line 66: Why this special case? Lists can be nullable or not. If they are, they should be marked as such.

Also, we are definitely not wrapping in `Optional` here. I think we should just ignore this option for Kotlin, unless it actually means using `java.util.Optional` *instead* of regular nullable types.

src/main/java/com/kobylynskyi/graphql/codegen/kotlin/KotlinGraphQLTypeMapper.java
```diff
@@ -0,0 +1,145 @@
...
+        if (Utils.isNotBlank(mappingContext.getApiReturnType())) {
+            // in case it is query/mutation and apiReturnType is set
+            return getGenericsString(mappingContext, mappingContext.getApiReturnType(), computedTypeName);
+        }
+        return getTypeConsideringPrimitive(mappingContext, namedDefinition, computedTypeName);
+    }
+
+    @Override
+    public boolean isPrimitive(String scalaType) {
+        return isKotlinPrimitive(scalaType);
```
joffrey-bion, line 86: Is this actually a reference to Scala (then why?) or did you mean *scalar*?

src/main/java/com/kobylynskyi/graphql/codegen/kotlin/KotlinGraphQLTypeMapper.java
```diff
@@ -0,0 +1,145 @@
...
+            case "Int":
+            case "Byte":
+            case "Short":
+            default:
+                return "0";
+        }
+    }
+
+    private String getOptionString(MappingContext mappingContext, String typeParameter) {
+        return typeParameter + KotlinGraphQLTypeMapper.KOTLIN_UTIL_OPTIONAL;
```
joffrey-bion, line 125: We should avoid the "optional" terminology here. Kotlin has nullable types and non-nullable types.

src/main/resources/templates/kotlin-lang/kotlinClassGraphqlResponseProjection.ftl
```diff
@@ -0,0 +1,94 @@
+<#if package?has_content>
+package ${package}
+
+</#if>
+import com.kobylynskyi.graphql.codegen.model.graphql.GraphQLResponseField
+import com.kobylynskyi.graphql.codegen.model.graphql.GraphQLResponseProjection
+<#if equalsAndHashCode>
+import java.util.Objects
```
joffrey-bion, line 8: Kotlin code is not necessarily for the JVM, so JDK classes will not be available if for instance we generate code for Kotlin/JS or Kotlin/Native.

src/main/resources/templates/kotlin-lang/kotlinClassGraphqlResponseProjection.ftl
```diff
@@ -0,0 +1,94 @@
...
+
+<#if javaDoc?has_content>
+/**
+<#list javaDoc as javaDocLine>
+ * ${javaDocLine}
+</#list>
+ */
+</#if>
+<#if generatedInfo.getGeneratedType()?has_content>
+@${generatedInfo.getGeneratedType()}(
```
joffrey-bion, line 19: We need to be careful with the default value of this for Kotlin generation.
I'm not sure there is a Kotlin equivalent to `javax.annotations.Generated`, so I guess by default we should put nothing, because Kotlin is not necessarily intended for the JVM.

src/main/resources/templates/kotlin-lang/kotlinClassGraphqlResponseProjection.ftl
joffrey-bion: Maybe we can deal with other Kotlin targets in a future PR.
There might be other things to consider to support other Kotlin targets.

src/main/java/com/kobylynskyi/graphql/codegen/kotlin/KotlinGraphQLTypeMapper.java
jxnu-liguobin: Options in the list are very common. I feel that this possibility needs to be preserved. As for whether to use it or not, it should be decided by the user's schema

src/main/resources/templates/kotlin-lang/kotlinClassGraphqlResponseProjection.ftl
jxnu-liguobin: At present,we should only consider the JVM, otherwise Scala also has Scala.js , Scala native, the implementation is more complex. @kobylynskyi What about you?

src/main/resources/templates/kotlin-lang/kotlinClassGraphqlResponseProjection.ftl
jxnu-liguobin: The premise of our library is the JVM language, otherwise it does not match the library name.
If we think that way, we won't be able to fully support any other language. Because many other JVM languages are actually cross platform, such as Scala and kotlin

src/main/java/com/kobylynskyi/graphql/codegen/kotlin/KotlinGraphQLTypeMapper.java
jxnu-liguobin: on the other hand, such as
```
type A {
    a:[Int]
    b:[Int!]
    c:[Int!]!
    d:[Int]!
}
```

All of them should be allowed because of the semantics of graphql. Scala, because there is no default non null setting, only handles primitive types, but kotlin can fully resolve the semantics of graphql for all types.
Otherwise, we will lose some semantics of graphql artificially.

src/main/resources/templates/kotlin-lang/kotlinClassGraphqlResponseProjection.ftl
kobylynskyi: I think for now it is ok to have a JVM version of Kotlin (also considering the fact that the name of the library is still graphql-**java**-codegen). So I'm 👌 with it.

kobylynskyi (changes requested):

src/main/resources/templates/kotlin-lang/kotlinClassGraphqlResponseProjection.ftl
jxnu-liguobin: Yes, otherwise I would like to continue to implement Python and Rust, But the underlying library is based on `graphql -Java`,

plugins/gradle/example-client/build.gradle
```diff
@@ -8,11 +8,13 @@ plugins {
     // use the latest available version:
     // https://plugins.gradle.org/plugin/io.github.kobylynskyi.graphql.codegen
     id "io.github.kobylynskyi.graphql.codegen" version "4.0.2-SNAPSHOT"
+    id "org.jetbrains.kotlin.jvm" version "1.3.71"
 }
 
 mainClassName = "io.github.kobylynskyi.order.Application"
 
 dependencies {
+    compile "org.jetbrains.kotlin:kotlin-stdlib-jdk8"
```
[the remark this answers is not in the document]
jxnu-liguobin: I have thought about it, but I think the example of gradle is too heavy. Unlike Scala example, I find that it uses spring and mongodb.So I can add the same (simple)example as SBT. This example only refers to the plug-in and the library itself. What do you think?

src/main/resources/templates/kotlin-lang/kotlinClassGraphqlResponseProjection.ftl
joffrey-bion: The name of the library is actually only Java, not the JVM. Bringing in more languages is already out of the scope of the initial library, so I'm not sure we can reason this way.

The question is whether the generated code actually needs library classes that are JVM only. While I think this is the case at the moment, I'm not sure it needs to be this way forever.
But I agree that if this is changed, the change has nothing to do with this particular PR, so it could wait.

plugins/gradle/example-client/build.gradle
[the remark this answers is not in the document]
joffrey-bion: People might be confused by the kotlin plugin and stdlib dependency here if they just want to use the lib for Java codegen. I also believe it would be much better to separate. I know it's a bit heavy but it would be worth it.
If you don't want to do it I think it would be less confusing and maybe better to have no Kotlin example at all rather than mix it with the Java one.

src/main/resources/templates/kotlin-lang/kotlinClassGraphqlResponseProjection.ftl
joffrey-bion: Agreed. At least for the server-side code generation (which was the initial intent of the lib) we are relying on graphql-java and I don't think this will change anytime soon.

src/main/java/com/kobylynskyi/graphql/codegen/kotlin/KotlinGraphQLTypeMapper.java
joffrey-bion: So it looks like the List case is handled later by `getTypeConsideringPrimitive()`, which also deals with appending `?` (but only if not already there).
I think there is definitely no need to handle this nullability here, and we should just let `getTypeConsideringPrimitive` deal with it.
The point of having a piece of code here (I think) is only if we want to given a different meaning to `useOptionalForNullableReturnTypes`, which I mentioned in my point 2.

src/test/java/com/kobylynskyi/graphql/codegen/scala/GraphQLCodegenExtendTest.java
```diff
@@ -37,7 +37,7 @@ void init() {
 
     @AfterEach
     void cleanup() {
-        Utils.deleteDir(outputBuildDir);
+//        Utils.deleteDir(outputBuildDir);
```
kobylynskyi, line 40: Did you commit it by accident?

kobylynskyi (changes requested):

src/main/java/com/kobylynskyi/graphql/codegen/kotlin/KotlinGraphQLTypeMapper.java
jxnu-liguobin: Sorry, the comment is directly modified when copying, which conflicts with the code later.

The optional function of Java is limited. If you consider that option and nullable are different, there may be```Option< type?>```, which I think is ambiguous. So I wonder if it is possible to temporarily not support ```useOptionalForNullableReturnTypes ``` ？(Just like Scala's option, the implementation in the template is very tedious due to the Java code in the template). At the same time, I will remove this special logic, and make sure that they have been handled by ```getTypeConsideringPrimitive ```.

src/main/resources/templates/kotlin-lang/kotlinClassGraphqlResponseProjection.ftl
jxnu-liguobin: I agree that clients may be used across platforms.

plugins/gradle/example-client/build.gradle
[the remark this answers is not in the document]
jxnu-liguobin: OK, I'll take the time to write an example and try the generated code.

src/test/java/com/kobylynskyi/graphql/codegen/scala/GraphQLCodegenExtendTest.java
jxnu-liguobin: I will remove it when PR is ready.

src/main/java/com/kobylynskyi/graphql/codegen/kotlin/KotlinGraphQLTypeMapper.java
joffrey-bion: > The optional function of Java is limited. If you consider that option and nullable are different, there may be `Option<type?>`, which I think is ambiguous.

There is no reason for `Optional<Type?>` to be generated. We control the generation. The only input we can have is a nullable or non-nullable graphql type. My point is that the `useOptionalForNullableReturnTypes` in the context of Kotlin should either be ignored (like you said) or simply make us generate `java.util.Optional<T>` *instead of* `T?`, but we would of course never generate `Optional<T?>`.

I agree we can simply ignore the option for now, but since it exists it doesn't cost much to just support it in the near future.

jxnu-liguobin: I have completed the compatibility of four situations.```wrapIntoList``` and```wrapSuperTypeIntoList``` can handle most cases, but with some exceptions, I implemented them in ```getTypeConsideringPrimitive```
```kotlin
@javax.annotation.Generated(
    value = ["com.kobylynskyi.graphql.codegen.GraphQLCodegen"],
    date = "2020-12-31T23:59:59-0500"
)
interface QueryResolver {

    @Throws(Exception::class)
    fun events(): List<Event>?

    @Throws(Exception::class)
    fun event(): Event?

    @Throws(Exception::class)
    fun null1Query(): Int?

    @Throws(Exception::class)
    fun null2Query(): List<Int?>?

    @Throws(Exception::class)
    fun null3Query(): List<Int>?

    @Throws(Exception::class)
    fun null4Query(): List<Int>

    @Throws(Exception::class)
    fun null5Query(): List<Int?>

}
```

Fortunately, it doesn't contain ```Optional```, otherwise I feel suffocated

joffrey-bion (approved):

kobylynskyi (approved):

kobylynskyi:
@jxnu-liguobin, the PR says "This pull request is still a work in progress".
Do you plan to push some more changes? Or I can merge this?

jxnu-liguobin:
> @jxnu-liguobin, the PR says "This pull request is still a work in progress".
> Do you plan to push some more changes? Or I can merge this?

I added an extreme test.At present, it looks good. 

```
type Query {
    events: [Event!]
    event: Event
    events1: [Event!]!
    events2: [Event]!
    events3: [Event]
    events4: [[Event]!]
    events5: [[Event]!]!
    events6: [[Event!]!]!
    events7: [[Event]]!
    ...
}
```
```kotlin
@javax.annotation.Generated(
    value = ["com.kobylynskyi.graphql.codegen.GraphQLCodegen"],
    date = "2020-12-31T23:59:59-0500"
)
interface QueryResolver {

    @Throws(Exception::class)
    fun events(): List<Event>?

    @Throws(Exception::class)
    fun event(): Event?

    @Throws(Exception::class)
    fun events1(): List<Event>

    @Throws(Exception::class)
    fun events2(): List<Event?>

    @Throws(Exception::class)
    fun events3(): List<Event?>?

    @Throws(Exception::class)
    fun events4(): List<List<Event?>>?

    @Throws(Exception::class)
    fun events5(): List<List<Event?>>

    @Throws(Exception::class)
    fun events6(): List<List<Event>>

    @Throws(Exception::class)
    fun events7(): List<List<Event?>?>
}
```

kobylynskyi (approved):

PR #471, opened 2020-12-28
support msg for deprecated

jxnu-liguobin:
support
```
directive @deprecated(reason: String = "No longer supported") on FIELD_DEFINITION | ENUM_VALUE
```

### DIFF
--- a/src/main/java/com/kobylynskyi/graphql/codegen/mapper/GraphQLTypeMapper.java
+++ b/src/main/java/com/kobylynskyi/graphql/codegen/mapper/GraphQLTypeMapper.java
@@ -319,11 +319,11 @@ public interface GraphQLTypeMapper {
     }
 
     default DeprecatedDefinition getDeprecated(MappingContext mappingContext, DirectivesContainer<?> directivesContainer) {
-        return directivesContainer.getDirectives().
-                stream().
-                filter(d -> d.getName().equalsIgnoreCase(Deprecated.class.getSimpleName())).
-                findFirst().
-                map(directive -> MultiLanguageDeprecated.getLanguageDeprecated(mappingContext.getGeneratedLanguage(), directive))
+        return directivesContainer.getDirectives()
+                .stream()
+                .filter(d -> d.getName().equalsIgnoreCase(Deprecated.class.getSimpleName()))
+                .findFirst()
+                .map(directive -> MultiLanguageDeprecated.getLanguageDeprecated(mappingContext.getGeneratedLanguage(), directive))
                 .orElse(null);
     }
 
--- a/src/main/java/com/kobylynskyi/graphql/codegen/mapper/GraphQLTypeMapper.java
+++ b/src/main/java/com/kobylynskyi/graphql/codegen/mapper/GraphQLTypeMapper.java
@@ -1,24 +1,11 @@
 package com.kobylynskyi.graphql.codegen.mapper;
 
-import com.kobylynskyi.graphql.codegen.model.GeneratedLanguage;
-import com.kobylynskyi.graphql.codegen.model.MappingContext;
-import com.kobylynskyi.graphql.codegen.model.NamedDefinition;
+import com.kobylynskyi.graphql.codegen.model.*;
 import com.kobylynskyi.graphql.codegen.model.definitions.ExtendedDefinition;
 import com.kobylynskyi.graphql.codegen.utils.Utils;
-import graphql.language.Argument;
-import graphql.language.Directive;
-import graphql.language.DirectivesContainer;
-import graphql.language.ListType;
-import graphql.language.NamedNode;
-import graphql.language.NonNullType;
-import graphql.language.Type;
-import graphql.language.TypeName;
+import graphql.language.*;
 
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
 
 /**
  * Map GraphQL type to language-specific type (java/scala/kotlin/etc)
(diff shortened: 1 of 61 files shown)


src/main/java/com/kobylynskyi/graphql/codegen/model/MultiLanguageDeprecated.java
```diff
@@ -0,0 +1,31 @@
...
+ * @version 1.0, 2020/12/28
+ */
+public class MultiLanguageDeprecated {
+
+    public static DeprecatedDefinition getLanguageDeprecated(GeneratedLanguage generatedLanguage, Directive directive) {
+        String JAVA = "Deprecated";
+        String SCALA = "deprecated";
+        String KOTLIN = "Deprecated";
+        String msg = null;
+        if (directive.getArguments().parallelStream().anyMatch(argument -> argument.getName().equals("reason"))) {
```
kobylynskyi, line 17: 1. Please remove `.parallelStream()` as it negatively impacts performance in our case.
2. Extract "reason" as constant: `private static final`

src/main/java/com/kobylynskyi/graphql/codegen/model/definitions/ExtendedFieldDefinition.java
```diff
@@ -24,10 +28,10 @@ protected ExtendedFieldDefinition(FieldDefinition f, boolean fromExtension) {
         this.fromExtension = fromExtension;
     }
 
-    public boolean isDeprecated() {
-        return getDirectives().stream()
-                .map(Directive::getName)
-                .anyMatch(Deprecated.class.getSimpleName()::equalsIgnoreCase);
+    public DeprecatedDefinition getDeprecated(MappingContext mappingContext) {
+        Optional<Directive> deprecatedDirective = getDirectives().parallelStream().
```
kobylynskyi, line 32: 1. Remove `parallelStream()`
2. Convert to a single-line:
```java
getDirectives()
    .filter(d -> d.getName().equalsIgnoreCase(Deprecated.class.getSimpleName()))
    .map(directive -> MultiLanguageDeprecated.getLanguageDeprecated(mappingContext.getGeneratedLanguage(), directive))
    .findFirst()
    .orElse(null);
```

kobylynskyi (changes requested):

src/main/java/com/kobylynskyi/graphql/codegen/model/MultiLanguageDeprecated.java
jxnu-liguobin: I don't get it? Is it ```msg```?

kobylynskyi: No, msg is fine. I am taking about parallel streams. They should be removed.

jxnu-liguobin: Is it because of the stream order? I've taken it all out

src/main/java/com/kobylynskyi/graphql/codegen/mapper/GraphQLTypeMapper.java
```diff
@@ -330,5 +318,14 @@ default String getTypeConsideringPrimitive(MappingContext mappingContext,
         return computedTypeName;
     }
 
+    default DeprecatedDefinition getDeprecated(MappingContext mappingContext, DirectivesContainer<?> directivesContainer) {
+        return directivesContainer.getDirectives().
```
kobylynskyi, line 322: Please not mix-up leading dots and trailing dots in the chained method calls.
(across the project, we are using the leading dot)

src/main/java/com/kobylynskyi/graphql/codegen/model/DeprecatedDefinition.java
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * @author 梦境迷离
+ * @version 1.0, 2020/12/28
+ */
+public class DeprecatedDefinition {
+
+    // For different languages, real implementation.
+    private final String annotation;
+    private final String reason;
+    private final static String DEFAULT_MSG = "No longer supported";
```
kobylynskyi, line 12: Minor note: constants are usually declared before fields

src/main/java/com/kobylynskyi/graphql/codegen/model/definitions/ExtendedFieldDefinition.java
```diff
@@ -24,10 +26,13 @@ protected ExtendedFieldDefinition(FieldDefinition f, boolean fromExtension) {
         this.fromExtension = fromExtension;
     }
 
-    public boolean isDeprecated() {
-        return getDirectives().stream()
-                .map(Directive::getName)
-                .anyMatch(Deprecated.class.getSimpleName()::equalsIgnoreCase);
+    public DeprecatedDefinition getDeprecated(MappingContext mappingContext) {
```
kobylynskyi, line 29: In this project, we are using trailing dots in the chained method calls.

src/main/java/com/kobylynskyi/graphql/codegen/model/MultiLanguageDeprecated.java
kobylynskyi: Thanks.
FYI: https://stackoverflow.com/questions/20375176/should-i-always-use-a-parallel-stream-when-possible

kobylynskyi (changes requested):
Good job!
I have just a few minor comments. If you can address them then we are good to go.

kobylynskyi (approved):

PR #477, opened 2021-01-09
Exception fix for non-query compound requests

WouterG:
Hi,

I encountered a bug while using the GraphQLRequests#toHttpJsonBody method, causing it to be unable to compile requests of non-query types.

I swapped the code to use the first request's operation type instead of using the default QUERY type, as the value would only be updated in the loop, after the check with exception had already happened.

The operation of getting the first entry is safe to use as the list is checked before to not be empty.

Hope to hear from you, and have an update for our usage 👍


kobylynskyi (approved):
@WouterG, looks good to me.
Thanks a lot for the contribution!
I will release it as part of 4.1.1 this weekend.

kobylynskyi:
@WouterG, version [4.1.1](https://github.com/kobylynskyi/graphql-java-codegen/releases/tag/v4.1.1) was released today.
Thanks again for the fix you provided.

Issue #479, opened 2021-01-12
[Scala] Consider generating additional classes required for serialization of scala enumerations


jxnu-liguobin:
**Is your feature request related to a problem? Please describe.**
How to serialize Scala enumeration correctly?
Such as graphql schema
```graphql
enum Episode {
    NEWHOPE
    EMPIRE
    JEDI
}

type Human implements Character {
    id: ID!
    name: String!
    friends: [Character]
    appearsIn: [Episode]!
    homePlanet: String
    secretBackstory : String @deprecated(reason : "We have decided that this is not canon")
    email: Email!
}
```
need follow config  to add into ```customAnnotationsMapping```
```scala
  
  mapping.put("Human.appearsIn", util.Arrays.asList("@com.fasterxml.jackson.module.scala.JsonScalaEnumeration(classOf[io.github.dreamylost.EpisodeEnum])"))
```
**Describe the solution you'd like**
Generate ```EpisodeEnum.scala```

Most commonly, when using Jackson operation scala enumeration, if ```Episode``` is an enumeration type, an ```EpisodeEnum``` must be generated for ```Episode```. 
```EpisodeEnum``` as follow
```scala
class EpisodeEnum extends TypeReference[Episode.type]// codegen will only generate  Episode.scala
```
Also this can be called **EpisodeEnumType**?
But it only works if Jackson is the serialization tool.So it's not sure if it needs to be done.

**Describe alternatives you've considered**

**Additional context**

kobylynskyi:
Will be released as part of v4.1.2

PR #483, opened 2021-01-14
support to generate xxTypeRefer for scala enum

jxnu-liguobin:
By the way, extract the version number of the test project and change readme to one example.
https://github.com/kobylynskyi/graphql-java-codegen/issues/479


src/main/resources/templates/scala-lang/scalaClassGraphqlEnum.ftl
```diff
@@ -2,6 +2,8 @@
 package ${package}
 
 </#if>
+import com.fasterxml.jackson.core.`type`.TypeReference
```
kobylynskyi, line 5: @jxnu-liguobin, do you need to wrap this import into `<#if serializationLibrary == 'JACKSON'>`?

kobylynskyi (changes requested):

src/main/resources/templates/scala-lang/scalaClassGraphqlEnum.ftl
jxnu-liguobin: yes, will change

src/test/resources/expected-classes/scala/deprecated/Status.scala.txt
```diff
@@ -1,5 +1,7 @@
 package com.github.graphql
 
+import com.fasterxml.jackson.core.`type`.TypeReference
```
kobylynskyi, line 3: Do you also need to do the “if JACKSON” check?

kobylynskyi (changes requested):

src/test/resources/expected-classes/scala/deprecated/Status.scala.txt
jxnu-liguobin: Only enumeration is needed, and only when Jackson is used

PR #557, opened 2021-02-25
Fix a case when field is non-null in type and nullable in interface #556

### LINKED ISSUE #556: Compilation error when field is non-null in type and nullable in interface [Version 4.1.0]
## Issue Description

If you have an interface that defines an Int (optional)
and you have a type that implements the interface but turns it to Required (Int!)

the resultant class that gets generated uses the primitive type `int` instead of the Object Integer.  All other int is Integer.  This seems like a bug?

The workaround currently is to add this in the customTypesMapping

mapping.put("Int!", "java.lang.Integer") <-- notice that you HAVE to specify Int! with the exclamation mark.  This could be another bug?

## Steps to Reproduce

Have your graphqls schema as follows:

type First implements Second {
  test: Int!
}

interface Second {
 test: Int
}

Run code generator with Java.

## Expected Result

First.java should be Integer test.

## Actual Result
*Explain in detail what behavior actually happened.*
First.java is primitive type int.

## Environment and Setup
* **graphql-java-codegen**: *4.1.0*
* **Build tool**: *E.g.: SBT*

kobylynskyi:
Fixes #556

### DIFF
--- a/src/main/java/com/kobylynskyi/graphql/codegen/mapper/FieldDefinitionToParameterMapper.java
+++ b/src/main/java/com/kobylynskyi/graphql/codegen/mapper/FieldDefinitionToParameterMapper.java
@@ -121,6 +121,7 @@ public class FieldDefinitionToParameterMapper {
         parameter.setAnnotations(graphQLTypeMapper.getAnnotations(mappingContext, fieldDef.getType(), fieldDef, parentTypeName, false));
         parameter.setJavaDoc(fieldDef.getJavaDoc());
         parameter.setDeprecated(fieldDef.getDeprecated(mappingContext));
+        parameter.setMandatory(namedDefinition.isMandatory());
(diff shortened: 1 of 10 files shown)


kobylynskyi:
@jxnu-liguobin do you need to do a similar thing for Kotlin/Scala?

jxnu-liguobin:
> @jxnu-liguobin do you need to do a similar thing for Kotlin/Scala?

Kotlin and Scala have no special primitive types, which I don't think need do it. Of course,I need to test it.

jxnu-liguobin:
Hi, @kobylynskyi 
Kotlin can override Int from Int? Unfortunately, but I found some new mistakes.

1. Scala cannot use Java `Object`'s method as case class' field name.(case class extends Object when compile time).
2. Kotlin toString generate error, when type is `String`.
3. Can not override from Option[Int] to Int in Scala, So, this problem cannot be solved, the only way is to use `mapping.put ("Int", "Int")`.

I have added test for them.

PR #567, opened 2021-03-04
Automate release process #562

### LINKED ISSUE #562: Automatically publish when the master is built successfully.
I guess the problem here is the change of the version number of the example projects.

I forgot to release 4.1.2 and 4.1.3. I just released 4.1.3. 😅

kobylynskyi:
Resolves #562

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -40,21 +40,21 @@ jobs:
 
       - name: Update version of Maven plugin to SNAPSHOT
         working-directory: plugins/maven/graphql-java-codegen-maven-plugin
-        run: mvn versions:set -DnewVersion="${{ github.event.inputs.release_type }}-SNAPSHOT" -DgenerateBackupPoms=false
+        run: mvn versions:set -DnewVersion="${{ github.event.inputs.release_version }}-SNAPSHOT" -DgenerateBackupPoms=false
 
       - name: Update version of Maven example client sub-project
         working-directory: plugins/maven/example-client
-        run: mvn versions:set -DnewVersion="${{ github.event.inputs.release_type }}" -DgenerateBackupPoms=false
+        run: mvn versions:set -DnewVersion="${{ github.event.inputs.release_version }}" -DgenerateBackupPoms=false
 
       - name: Update version of Maven example server sub-project
         working-directory: plugins/maven/example-server
-        run: mvn versions:set -DnewVersion="${{ github.event.inputs.release_type }}" -DgenerateBackupPoms=false
+        run: mvn versions:set -DnewVersion="${{ github.event.inputs.release_version }}" -DgenerateBackupPoms=false
 
       - name: Update versions of all Gradle/SBT sub-projects
-        run: scripts/update-release-version-in-build-files.sh ${{ github.event.inputs.release_type }}
+        run: scripts/update-release-version-in-build-files.sh ${{ github.event.inputs.release_version }}
 
       - name: Update versions in all README files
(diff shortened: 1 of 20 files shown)


kobylynskyi:
@jxnu-liguobin please take a look

jxnu-liguobin:
> @jxnu-liguobin please take a look

Ok

.github/workflows/release.yml
```diff
@@ -0,0 +1,146 @@
...
+        run: |
+          mvn \
+            --no-transfer-progress \
+            --batch-mode \
+            -Dgpg.passphrase=${{ secrets.OSSRH_GPG_SECRET_KEY_PASSWORD }} \
+            release:clean release:prepare release:perform
+
+      - name: Release SBT plugin
+        working-directory: plugins/sbt/graphql-java-codegen-sbt-plugin
+        run: sbt release with-defaults default-tag-exists-answer k
```
jxnu-liguobin, line 138: This will use the default version, such as (3.1.1 - > 3.1.2-SNAPSHOT), But other plugins use `${{ github.event.inputs.release_version }}`, This may need to change.

Like this, We can use args  `release-version 1.0.99 next-version 1.2.0-SNAPSHOT`
image

kobylynskyi: @jxnu-liguobin this workflow will always be executed from the master branch which will not have any SHANPSHOT versions.

jxnu-liguobin: > @jxnu-liguobin this workflow will always be executed from the master branch which will not have any SHANPSHOT versions.

Will the next version be updated after the release?

.github/workflows/release.yml
```diff
@@ -0,0 +1,146 @@
...
+
+
+      - name: Commit and push the new version
```
jxnu-liguobin, line 141: I guess the script will remove SNAPSHOT from all files before publishing, and the next version to be published will be submitted after publishing, but I wonder if gradle and Maven will automatically change to the next version after release? Because SBT will. So, What does `new version` in `Commit and push the new version` mean?

.github/workflows/release.yml
kobylynskyi: The idea is the following:
In **master branch** (from which the release will be made) - we will keep the original release version and do not increment to the next SNAPSHOT version.
In **develop branch** (after release will be made) - we will increment to the next SNAPSHOT version.

.github/workflows/release.yml
kobylynskyi: maven will automatically increment the version **only for the plugin**, but not the example projects.
gradle **will not** automatically increment the version at all (plugin + example projects)
library version will also **not** be incremented automatically.

So I will update versions using the script I wrote in this PR before commit/push.

.github/workflows/release.yml
jxnu-liguobin: very good! 👍

.github/workflows/release.yml
jxnu-liguobin: Everything looks good, but I'm afraid there might be something wrong with `with-defaults` in Non interactive environment. And OSS account has been sent to your Google email. If there is any problem in the test, please inform me to change the SBT task at any time.

kobylynskyi: I didn’t receive any email yet.. did you send it to 92bogdan@gmail.com?

jxnu-liguobin: yes

jxnu-liguobin: I used outlook mailbox, garbage can to see?

jxnu-liguobin (dismissed):

PR #576, opened 2021-03-07
fix enum import.

jxnu-liguobin:
This is because the type may be a `Seq[Enum].`


kobylynskyi (approved):

kobylynskyi:
Do we have a corresponding issue for this in Issues section?

jxnu-liguobin:
> Do we have a corresponding issue for this in Issues section?

No, I just found it using the GitHub API.

PR #607, opened 2021-03-22
Fix default kotlin double mapping value (#600)

### LINKED ISSUE #600: Default Kotlin mapped value for Double is not supported
Hello,

We are declaring a property to be a non-nullable `Float`:

```
type TheThing {
  someString: String!
  someFloat: Float!
}
```

And the generated Kotlin class is:
```
package mil.kr.direct.graphql.model


@javax.annotation.processing.Generated(
    value = ["com.kobylynskyi.graphql.codegen.GraphQLCodegen"],
    date = "2021-03-19T15:05:41-0400"
)
data class TheThing(
    var someString: String,
    var someFloat: Double
) {

    companion object {
        fun builder(): Builder = Builder()
    }


    class Builder {

        private lateinit var someString: String
        private var someFloat: Double = 0D

        fun setSomeString(someString: String): Builder {
            this.someString = someString
            return this
        }

        fun setSomeFloat(someFloat: Double): Builder {
            this.someFloat = someFloat
            return this
        }

        fun build(): TheThing {
            return TheThing(someString, someFloat)
        }
    }
}
```

However, the line `private var someFloat: Double = 0D` is not compiling as valid Kotlin:
```
Unsupported [literal prefixes and suffixes]
Unresolved reference: D
```

It expects the value to be `0.0` instead of `0D`. 

I'm guessing this is the source of that mapping?

https://github.com/kobylynskyi/graphql-java-codegen/blob/e306361fa0f6a2e755ee49b9fbaa306c99ae2151/src/main/java/com/kobylynskyi/graphql/codegen/kotlin/KotlinGraphQLTypeMapper.java#L52-L53

Is there something on our end (like a configuration) that I'm missing? Otherwise, could this be amended? I could also make a PR for it if it's just this method change.

Thank you!

aigyptios:
This fixes #600

### DIFF
--- a/src/main/java/com/kobylynskyi/graphql/codegen/kotlin/KotlinGraphQLTypeMapper.java
+++ b/src/main/java/com/kobylynskyi/graphql/codegen/kotlin/KotlinGraphQLTypeMapper.java
@@ -50,7 +50,7 @@ public class KotlinGraphQLTypeMapper implements GraphQLTypeMapper {
             case "Float":
                 return "0F";
             case "Double":
-                return "0D";
+                return "0.0";
             case "Char":
                 return "0.toChar()";
             case "Boolean":


kobylynskyi (approved):
@jxnu-liguobin could you please review?

jxnu-liguobin:
> @jxnu-liguobin could you please review?

no problem. I have tested it.

Issue #606, opened 2021-03-21
[Feature Request] Add ability to generate particular types as Java interface, not POJO class


donbeave:
**Is your feature request related to a problem? Please describe.**

Let imagine we have a GraphQL service with such schema:

```graphql
type Query {
  userCurrent: User
}

type User {
  username: String!
  email: String!
  orders: [Order!]! @customResolver
}

type Order {
  number: String!
  price: String!
}
```

And in the database, we store `users` and `orders` in separate tables. We have such JPA models and repositories:

```java
@Entity
public class UserJPA {
    @Id
    @GeneratedValue(strategy = GenerationType.IDENTITY)
    private Long id;

    @NotNull
    private String userName;

    @NotNull
    private String emailAddress;

    // getters and setters
}
```

```java
@Entity
public class OrderJPA {
    @Id
    @GeneratedValue(strategy = GenerationType.IDENTITY)
    private Long id;

    @ManyToOne(fetch = FetchType.LAZY)
    @JoinColumn
    private User user;

    @NotNull
    private String number;

    @NotNull
    private BigDecimal price = BigDecimal.ZERO;

    // getters and setters
}
```

```java
@Repository
public interface OrderRepository extends JpaRepository<OrderJPA, Long> {

    List<OrderJPA> findAllByUser(UserJPA user);

}
```

In the GraphQL server, we want to load orders only when they were requested by the client.

1) this query will NOT load orders from the database:

```graphql
query {
  userCurrent {
    username
    email
  }
}
```

1) but this query will load orders for the current user from the database:

```graphql
query {
  userCurrent {
    username
    orders {
      number
    }
  }
}
```

Now when we use `graphql-java-codegen` to generate a java code we will have something like this:

models:

```java
public class User implements java.io.Serializable {
    private String username;
    private String email;

    public String getUsername() {
        return username;
    }
    public void setUsername(String username) {
        this.username = username;
    }

    public String getEmail() {
        return email;
    }
    public void setEmail(String email) {
        this.email = email;
    }
}
```

```java
public class Order implements java.io.Serializable {
    private String number;

    public String getNumber() {
        return number;
    }
    public void setNumber(String number) {
        this.number = number;
    }
}
```

api:

```java
public interface UserCurrentQueryResolver extends graphql.kickstart.tools.GraphQLQueryResolver {
    User userCurrent(graphql.schema.DataFetchingEnvironment env) throws Exception;
}
```

```java
[...]

PR #614, opened 2021-03-26
Support typesAsInterfaces in Gradle and SBT plugins #606

kobylynskyi:
CC: @donbeave @jxnu-liguobin


donbeave:
@kobylynskyi thank you!

Issue #626, opened 2021-04-06
 [Feature Request] Add ability to generate union type resolver based on __typename field


gdhordain:
**Is your feature request related to a problem? Please describe.**
When using union type with graphql-java-codegen, we need to define a way to tell Jackson how to choose which implementation depending on the data.
This step could be generated.

**Describe the solution you'd like**
The spec (https://graphql.org/learn/schema/#union-types) defines the `__typename` field which will be populated with the data type.
It would be possible to generate Jackson annotations & resolver based on the `__typename` field.
This behaviour could be linked to a configuration toggle, and could be overwritten by `customAnnotationsMapping` configuration.

**Describe alternatives you've considered**
Stay with current implementation.

jxnu-liguobin:
If I understand it well, do you want to generate these annotations directly to the interface/class?
```
customAnnotationsMapping := {
  val mapping = new util.HashMap[String, util.List[String]]
  val annotations = new util.ArrayList[String]()
  annotations.add("@com.fasterxml.jackson.annotation.JsonTypeInfo(use=com.fasterxml.jackson.annotation.JsonTypeInfo.Id.NAME, include=com.fasterxml.jackson.annotation.JsonTypeInfo.As.PROPERTY,property = \"__typename\")")
  annotations.add(
    """@com.fasterxml.jackson.annotation.JsonSubTypes(value = Array(
      |        new com.fasterxml.jackson.annotation.JsonSubTypes.Type(value = classOf[HumanDO], name = "Human"),
      |        new com.fasterxml.jackson.annotation.JsonSubTypes.Type(value = classOf[DroidDO], name = "Droid")))""".stripMargin)
  mapping.put("Character", annotations)
  mapping.put("Droid.appearsIn", util.Arrays.asList("@com.fasterxml.jackson.module.scala.JsonScalaEnumeration(classOf[io.github.dreamylost.model.EpisodeDOTypeRefer])"))
  mapping.put("Human.appearsIn", util.Arrays.asList("@com.fasterxml.jackson.module.scala.JsonScalaEnumeration(classOf[io.github.dreamylost.model.EpisodeDOTypeRefer])"))
  mapping
}
```
This demand is very common, and I have considered it. The annotation is not generated actively because there may be many serialization ways in Java. So, currently, the library only generates the tool class for serialization of Scala enumeration types(jackson).

[...]

PR #637, opened 2021-04-09
Generate jackson type id resolver

gdhordain:
Linked to issue #626.

Current status:
- [x] generate annotation
- [x] generate java resolver
- [x] generate kotlin resolver
- [x] generate scala resolver
- [x] plug with gradle plugin
- [x] plug with maven plugin
- [x] plug with sbt plugin


src/main/resources/templates/java-lang/javaClassGraphqlJacksonTypeIdResolver.ftl
```diff
@@ -0,0 +1,55 @@
...
+
+    @Override
+    public JavaType typeFromId(DatabindContext context, String typename) {
+        try {
+            var clazz = Class.forName(
+                <#if package?has_content>"${package}." +
+                </#if><#if modelNamePrefix?has_content>"${modelNamePrefix}" +
+                </#if>typename<#if modelNamePrefix?has_content> +
+                "${modelNameSuffix}"</#if>
+            );
```
kobylynskyi, line 33: @gdhordain, sorry, I didn't get why do we need to consider `modelNameSuffix`/`modelNamePrefix` here?

kobylynskyi (approved):

kobylynskyi:
@jxnu-liguobin could you please contribute kotlin and scala + sbt-related changes?
Thanks!

src/test/resources/expected-classes/jackson-resolver-union/GraphqlJacksonTypeIdResolver.java.txt
```diff
@@ -0,0 +1,49 @@
...
+
+    @Override
+    public void init(JavaType baseType) {
+        superType = baseType;
+    }
+
+    @Override
+    public JavaType typeFromId(DatabindContext context, String typename) {
+        try {
+            var clazz = Class.forName(
```
jxnu-liguobin, line 24: can it use `var` here  in java 8?

jxnu-liguobin:
@kobylynskyi @gdhordain 
Because both union and interface are required `__typename`, Also need generat jackson type id for `interface`.

jxnu-liguobin (changes requested):

src/main/java/com/kobylynskyi/graphql/codegen/mapper/GraphQLTypeMapper.java
```diff
@@ -312,6 +314,17 @@ default NamedDefinition getLanguageType(MappingContext mappingContext, String gr
             if (!Utils.isEmpty(annotationsToAdd)) {
                 annotations.addAll(annotationsToAdd);
             }
+        } else if (Boolean.TRUE.equals(mappingContext.getGenerateJacksonTypeIdResolver()) && isUnion) {
```
jxnu-liguobin, line 317: interface need

jxnu-liguobin:
> @jxnu-liguobin could you please contribute kotlin and scala + sbt-related changes?
> Thanks!

Yes, But It will take some time.

src/main/java/com/kobylynskyi/graphql/codegen/mapper/GraphQLTypeMapper.java
```diff
@@ -312,6 +314,17 @@ default NamedDefinition getLanguageType(MappingContext mappingContext, String gr
...
+            annotations.add("com.fasterxml.jackson.annotation.JsonTypeInfo(use = " +
+                    "com.fasterxml.jackson.annotation.JsonTypeInfo.Id.NAME, property = \"__typename\")");
+            String modelPackageName = DataModelMapper.getModelPackageName(mappingContext);
+            if (modelPackageName == null) {
+                modelPackageName = "";
+            } else if (Utils.isNotBlank(modelPackageName)) {
+                modelPackageName += ".";
+            }
+            annotations.add("com.fasterxml.jackson.databind.annotation.JsonTypeIdResolver(" + modelPackageName +
```
jxnu-liguobin, line 326: `add annotation` , It's not universal logic. It can't be put here.
Need to put in `JavaGraphqlTypeMapper`


```
 default List<String>  getExtractAnnotations(MappingContext mappingContext, boolean isUnion) {
        return new ArrayList<>();
    }

annotations.addAll(getExtractAnnotations(mappingContext,isUnion));
```


In JavaGraphQLTypeMapper
```
    public List<String> getExtractAnnotations(MappingContext mappingContext, boolean isUnion) {
        List<String> defaults = new ArrayList<>();
        if (Boolean.TRUE.equals(mappingContext.getGenerateJacksonTypeIdResolver()) && isUnion) {
            defaults.add("com.fasterxml.jackson.annotation.JsonTypeInfo(use = " +
                    "com.fasterxml.jackson.annotation.JsonTypeInfo.Id.NAME, property = \"__typename\")");
            String modelPackageName = DataModelMapper.getModelPackageName(mappingContext);
            if (modelPackageName == null) {
                modelPackageName = "";
            } else if (Utils.isNotBlank(modelPackageName)) {
                modelPackageName += ".";
            }
            defaults.add("com.fasterxml.jackson.databind.annotation.JsonTypeIdResolver(" + modelPackageName +
                    "GraphqlJacksonTypeIdResolver.class)");
        }
        return defaults;
    }
```

In Kotlin
```
    public List<String> getExtractAnnotations(MappingContext mappingContext, boolean isUnion) {
        List<String> defaults = new ArrayList<>();
        if (Boolean.TRUE.equals(mappingContext.getGenerateJacksonTypeIdResolver()) && isUnion) {
            defaults.add("com.fasterxml.jackson.annotation.JsonTypeInfo(use = " +
                    "com.fasterxml.jackson.annotation.JsonTypeInfo.Id.NAME, property = \"__typename\")");
            String modelPackageName = DataModelMapper.getModelPackageName(mappingContext);
            if (modelPackageName == null) {
                modelPackageName = "";
            } else if (Utils.isNotBlank(modelPackageName)) {
                modelPackageName += ".";
            }
            defaults.add("com.fasterxml.jackson.databind.annotation.JsonTypeIdResolver(" + modelPackageName +
                    "GraphqlJacksonTypeIdResolver::class)");
        }
        return defaults;
    }
```

jxnu-liguobin (changes requested):

src/test/resources/expected-classes/jackson-resolver-union/GraphqlJacksonTypeIdResolver.java.txt
kobylynskyi: No, in Java 8 it is not possible

jxnu-liguobin: I have update in my pr.

jxnu-liguobin:
I've finished all the work, just review https://github.com/MOOM-Planet/graphql-java-codegen/pull/1.

In order to complete PR faster, the annotation on the interface has not been solved for the time being.

> Of course, the PR can choose to solve only union. But I think it's a similar problem. It's better to solve it together as much as possible.

src/main/resources/templates/java-lang/javaClassGraphqlJacksonTypeIdResolver.ftl
gdhordain: If the model name prefix and/or suffix is used, then the generated class name will use them whereas the graphql typename value will remains untouched.

In my test, typename is `UnionMemberA`, generated class name is `MyUnionMemberASuffix`.

src/test/resources/expected-classes/jackson-resolver-union/GraphqlJacksonTypeIdResolver.java.txt
gdhordain: Yes, good catch.
My env was using Java11 at first, then I switched back to Java 8.

src/main/java/com/kobylynskyi/graphql/codegen/mapper/GraphQLTypeMapper.java
gdhordain: I had a look to your PR and merged it. Thanks !

jxnu-liguobin: A single test error was found and PR was submitted.

src/main/java/com/kobylynskyi/graphql/codegen/java/JavaGraphQLTypeMapper.java
```diff
@@ -122,4 +124,22 @@ public ValueMapper getValueMapper() {
...
+            String modelPackageName = DataModelMapper.getModelPackageName(mappingContext);
+            if (modelPackageName == null) {
+                modelPackageName = "";
+            } else if (Utils.isNotBlank(modelPackageName)) {
+                modelPackageName += ".";
+            }
+            defaults.add("com.fasterxml.jackson.databind.annotation.JsonTypeIdResolver(" + modelPackageName +
+                    "GraphqlJacksonTypeIdResolver.class)");
+        }
+        return defaults;
```
kobylynskyi, line 142: @gdhordain
There are around 10 lines of duplicate code in 3 classes (for each language)
So I propose to create an abstract method only for returning a JsonTypeIdResolver (so far I see there's a difference inly in this line:
```java
@Override
public List<String> getJsonTypeIdResolverAnnotationForPackage(String modePackageName) {
    return "com.fasterxml.jackson.databind.annotation.JsonTypeIdResolver("
        + modelPackageName + "GraphqlJacksonTypeIdResolver.class)";
}
```
Please let me know what you think. Thanks.

kobylynskyi (changes requested):

src/test/resources/expected-classes/jackson-resolver-union/GraphqlJacksonTypeIdResolver.java.txt
kobylynskyi: Good catch indeed. Thanks for fixing this.

src/main/java/com/kobylynskyi/graphql/codegen/java/JavaGraphQLTypeMapper.java
jxnu-liguobin: Yes, I didn't think too much about it at the weekend.

gdhordain: I pushed a refactor to avoid these duplicated lines.

jxnu-liguobin (approved):

src/main/java/com/kobylynskyi/graphql/codegen/java/JavaGraphQLTypeMapper.java
```diff
@@ -8,7 +8,9 @@
 import com.kobylynskyi.graphql.codegen.model.graphql.GraphQLOperation;
 import com.kobylynskyi.graphql.codegen.utils.Utils;
 
+import java.util.ArrayList;
```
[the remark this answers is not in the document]
kobylynskyi: @gdhordain please fix this minor issue. Thanks!

kobylynskyi (changes requested):

jxnu-liguobin:
Looks Good. 👍

src/main/java/com/kobylynskyi/graphql/codegen/mapper/GraphQLTypeMapper.java
```diff
@@ -285,10 +326,13 @@ default NamedDefinition getLanguageType(MappingContext mappingContext, String gr
      * @param parentTypeName  Name of the parent type
      * @param directives      List of GraphQL directive
      * @param mandatory       Type is mandatory
+     * @param isUnion         Is it union type
+     * @param typeName        Name of GraphQL type, but only not null when type is Enum
      * @return list of Java annotations for a given GraphQL type
      */
     default List<String> getAnnotations(MappingContext mappingContext, String graphQLTypeName, String name,
```
kobylynskyi, line 333: @gdhordain @jxnu-liguobin Instead of providing 2 extra parameters I'd recommend passing the type itself, so that required information can be fetched in the place it is needed. For example, you can pass either `Type<?>`, or `ExtendedDefinition<?, ?>` or `NamedDefinition` which then can be used in: 
* `ScalaGraphQLTypeMapper` (the only place `typeName` is used)
* `getJacksonTypeIdAnnotations` (the only place where `isUnion` is used)

Another benefit from passing a single type parameter is that in the future we might need to fetch another type attribute directly in the method instead of extending this method signature.

Please let me know what you think.
Hope it make sense to you.

kobylynskyi (changes requested):

src/main/java/com/kobylynskyi/graphql/codegen/mapper/GraphQLTypeMapper.java
jxnu-liguobin: It's good.

gdhordain: I makes sense, I'll do it.

gdhordain: I pushed a refactor which use the definition.
Is it ok for you ?

kobylynskyi (approved):
@gdhordain Looks great. Thanks!

PR #645, opened 2021-04-12
Ability to disable generation of the all$ method in projections

robbertnoordzij:
Hi,

Thanks for this great project. We are very fond of this project and use it for generating clients and models for our GraphQL layer. Last week, we had a small problem where a consumer started consuming more fields after the client was updated. The consumer used the all$ method in a projection. This resulted in some weird behavior and unnecessary calculations on the server.

We would like to be able to disable the generation of the all$ method in the projection, so that the consumer has to be explicit about the fields it uses.

In this pull request, I added an extra option: `generateAllMethodInProjection`. By default, I left the setting on true so it will not break any other projects. 

Let me know what you think about the approach and where I can improve it.

---

Changes were made to:
- [x] Codegen library - Java
- [x] Codegen library - Scala
- [x] Maven plugin
- [x] Gradle plugin
- [x] SBT plugin


robbertnoordzij:
I seem to get 403 from the repository during the Scala build. Is that normal or is there something wrong with the plugin?

kobylynskyi:
@robbertnoordzij will you be also able to make the changes to Maven, Gradle, and SBT plugin code? Thanks.

robbertnoordzij:
> @robbertnoordzij will you be also able to make the changes to Maven, Gradle, and SBT plugin code? Thanks.

I am not sure what I missed, the maven plugin is functioning. However I found out that while generating our local project with the flag set to false, maven will shout that the abstract `all$` method is not implemented from the `GraphQLResponseProjection` abstract class. I'm not sure were it is being used, I am able to remove it from the interface, the test cases and build the plugins and our project just fine.

Are you oke with removing the `all$` method from the abstract class? Or would you like me to create a separate interface that would be included when the option is enabled? This would make it a breaking change, I presume?

jxnu-liguobin:
Consider providing a default implementation for illegal, returning null or exception?

robbertnoordzij:
It can be deleted from the abstract, at least it seems not being used internally or in the generated code. In the generated code it will always call the concrete projection (example from a test):

```java
    public EventResponseProjection all$(int maxDepth) {
        this.status();
        this.createdDateTime();
        if (projectionDepthOnFields.getOrDefault("EventResponseProjection.AssetResponseProjection.assets", 0) <= maxDepth) {
            projectionDepthOnFields.put("EventResponseProjection.AssetResponseProjection.assets", projectionDepthOnFields.getOrDefault("EventResponseProjection.AssetResponseProjection.assets", 0) + 1);
            this.assets(new AssetResponseProjection().all$(maxDepth - projectionDepthOnFields.getOrDefault("EventResponseProjection.AssetResponseProjection.assets", 0)));
        }
        this.id();
        this.createdBy();
        this.typename();
        return this;
    }
```

Would you add the default empty implementation to the `GraphQLResponseProjection` or to the generated code. I would like the compiler to fail when the all$ is being used and not during runtime, that is why I am more in favour of removing it from the parent class.

I would suggest adding an extra marker interface (naming is not my strongest):

```java
interface GraphQLAllFieldsBehaviour {

    public GraphQLResponseProjection all$();

    public GraphQLResponseProjection all$(int maxDepth);
    
}
```

jxnu-liguobin:
> It can be deleted from the abstract, at least it seems not being used internally or in the generated code. In the generated code it will always call the concrete projection (example from a test):
> 
> ```java
>     public EventResponseProjection all$(int maxDepth) {
>         this.status();
>         this.createdDateTime();
>         if (projectionDepthOnFields.getOrDefault("EventResponseProjection.AssetResponseProjection.assets", 0) <= maxDepth) {
>             projectionDepthOnFields.put("EventResponseProjection.AssetResponseProjection.assets", projectionDepthOnFields.getOrDefault("EventResponseProjection.AssetResponseProjection.assets", 0) + 1);
>             this.assets(new AssetResponseProjection().all$(maxDepth - projectionDepthOnFields.getOrDefault("EventResponseProjection.AssetResponseProjection.assets", 0)));
>         }
>         this.id();
>         this.createdBy();
>         this.typename();
>         return this;
>     }
> ```
> 
> Would you add the default empty implementation to the `GraphQLResponseProjection` or to the generated code? I would like the compiler to fail when the all$ is being used and not during runtime, that is why I am more in favour of removing it from the parent class.
> 
> I would suggest adding an extra marker interface (naming is not my strongest):
> 
> ```java
> interface GraphQLAllFieldsBehaviour {
> 
>     public GraphQLResponseProjection all$();
> 
>     public GraphQLResponseProjection all$(int maxDepth);
>     
> }
> ```

It's feasible to extract upward. I think it's good. Note that `projectionDepthOnFields` must be extracted at the same time, because `all$` is recursive.

robbertnoordzij:
Extracting it upwards doesn't make it removable downwards right? Or do I misunderstand your suggestion.

The recursion is only taking place on the concrete implementation of the class. The `projectionDepthOnFields` could also be moved to the concrete implementation and be removed from the `GraphQLProjection` as well. The only thing remainging will be the toString method. 

We could also make the `GraphQLProjection` part of the generated source code and generate it based on the settings. But that would be an big change.

jxnu-liguobin:
> Extracting it upwards doesn't make it removable downwards right? Or do I misunderstand your suggestion.
> 
> The recursion is only taking place on the concrete implementation of the class. The `projectionDepthOnFields` could also be moved to the concrete implementation and be removed from the `GraphQLProjection` as well. The only thing remainging will be the toString method.
> 
> We could also make the `GraphQLProjection` part of the generated source code and generate it based on the settings. But that would be an big change.

Your solution is OK, because `projectionDepthOnFields` will only be used by `all`. It's meaningless to leave it in subclasses.

robbertnoordzij:
So I change the templates:

```ftl
</#if>
import com.kobylynskyi.graphql.codegen.model.graphql.GraphQLResponseField;
import com.kobylynskyi.graphql.codegen.model.graphql.GraphQLResponseProjection;
<#if fields?has_content && generateAllMethodInProjection>
import java.util.HashMap;
import java.util.Map;
</#if>
<#if equalsAndHashCode>
import java.util.Objects;
</#if>

(...)

public class ${className} extends GraphQLResponseProjection {
<#if fields?has_content && generateAllMethodInProjection>

    private final Map<String, Integer> projectionDepthOnFields = new HashMap<>();
</#if>

    public ${className}() {
    }
<#if fields?has_content && generateAllMethodInProjection>

    public ${className} all$() {
        return all$(${responseProjectionMaxDepth});
    }

    public ${className} all$(int maxDepth) {
    <#list fields as field>
        <#if field.type?has_content>
            <#if field.methodName?substring(0,2) != "on">
        if (projectionDepthOnFields.getOrDefault("${className}.${field.type}.${field.methodName}", 0) <= maxDepth) {
            projectionDepthOnFields.put("${className}.${field.type}.${field.methodName}", projectionDepthOnFields.getOrDefault("${className}.${field.type}.${field.methodName}", 0) + 1);
            this.${field.methodName}(new ${field.type}().all$(maxDepth - projectionDepthOnFields.getOrDefault("${className}.${field.type}.${field.methodName}", 0)));
        }
        </#if>
    <#else>
        this.${field.methodName}();
        </#if>
    </#list>
        return this;
    }
</#if>
```

This results in the following:

```java
   public class EventResponseProjection extends GraphQLResponseProjection {

        private final Map<String, Integer> projectionDepthOnFields = new HashMap<>();

        public EventResponseProjection() {
        }

        public EventResponseProjection all$() {
            return all$(3);
        }

        public EventResponseProjection all$(int maxDepth) {
            this.status();
            this.createdDateTime();
            if (projectionDepthOnFields.getOrDefault("EventResponseProjection.AssetResponseProjection.assets", 0) <= maxDepth) {
                projectionDepthOnFields.put("EventResponseProjection.AssetResponseProjection.assets", projectionDepthOnFields.getOrDefault("EventResponseProjection.AssetResponseProjection.assets", 0) + 1);
                this.assets(new AssetResponseProjection().all$(maxDepth - projectionDepthOnFields.getOrDefault("EventResponseProjection.AssetResponseProjection.assets", 0)));
            }
            this.id();
            this.createdBy();
            this.typename();
            return this;
        }
```

And then I will be able to remove the all$ method en protected field from the GraphQLResponseProjection.

If also @kobylynskyi is in favor of this approach, then I can change all the test assertion. (Because a lot are failing with this change :laughing: )

jxnu-liguobin:
It's just me. I like to use this way.
```java

interface GraphQLAllFieldsBehaviour {
     Map<String, Integer> projectionDepthOnFields = new HashMap<>();
     GraphQLResponseProjection all$();
     GraphQLResponseProjection all$(int maxDepth);
}
public class ${className} extends GraphQLResponseProjection implements GraphQLAllFieldsBehaviour {
     
    @Override
    public  GraphQLResponseProjection all$();
     @Override
    public GraphQLResponseProjection all$(int maxDepth);
}

public class ${className} extends GraphQLResponseProjection {

}
```

robbertnoordzij:
In your example `Map<String, Integer> projectionDepthOnFields = new HashMap<>();` in your example is an static e.g. single instance. Won't this create issues with race conditions while using this projection in threads and the same projections on different levels?

robbertnoordzij:
Then I would like to maintain it within the scope of the class how the depth is calculated.

```java
public class EventResponseProjection extends GraphQLResponseProjection {

        private final Map<String, Integer> projectionDepthOnFields = new HashMap<>();

        public EventResponseProjection() {
        }

        public EventResponseProjection all$() {
            return all$(3);
        }

        public EventResponseProjection all$(int maxDepth) {
            this.status();
            this.createdDateTime();
            if (projectionDepthOnFields.getOrDefault("EventResponseProjection.AssetResponseProjection.assets", 0) <= maxDepth) {
                projectionDepthOnFields.put("EventResponseProjection.AssetResponseProjection.assets", projectionDepthOnFields.getOrDefault("EventResponseProjection.AssetResponseProjection.assets", 0) + 1);
                this.assets(new AssetResponseProjection().all$(maxDepth - projectionDepthOnFields.getOrDefault("EventResponseProjection.AssetResponseProjection.assets", 0)));
            }
            this.id();
            this.createdBy();
            this.typename();
            return this;
        }
}
```

kobylynskyi:
What if we include this change in the next major release (6.0.0) and by default do not generate the $all method? I am in favor not including it by default because it confronts with GraphQL principle (however, can be used by clients to simplify the development/maintenance)

robbertnoordzij:
@kobylynskyi that whould be great, how would you than generate the all$ method? Would you use create `projectionDepthOnFields` within the projection class?

robbertnoordzij:
In this last commit I moved the `projectionDepthOnFields`. I tested the java implementation with our internal GraphQL schema and it works. I validated the scala output and kotlin output to make sure that these are still valid.

src/main/resources/templates/kotlin-lang/kotlinClassGraphqlResponseProjection.ftl
```diff
@@ -26,10 +26,12 @@ import java.util.Objects
 </#list>
 open class ${className} : GraphQLResponseProjection() {
 
-<#if fields?has_content>
-    override fun `all$`(): ${className} = `all$`(${responseProjectionMaxDepth})
+<#if fields?has_content && generateAllMethodInProjection>
+    var projectionDepthOnFields: MutableMap<String, Int> = HashMap<String, Int>()
```
jxnu-liguobin, line 30: Must use `MutableMap`, otherwise, it cannot be modified.
```kotlin

    val projectionDepthOnFields: MutableMap<String, Int> = mutableMapOf()
    projectionDepthOnFields.put(key, value)
```

src/main/resources/templates/scala-lang/scalaClassGraphqlResponseProjection.ftl
```diff
@@ -7,6 +7,9 @@ import com.kobylynskyi.graphql.codegen.model.graphql.GraphQLResponseProjection
 <#if equalsAndHashCode>
 import java.util.Objects
 </#if>
+<#if fields?has_content && generateAllMethodInProjection>
+import scala.collection.mutable.HashMap
```
jxnu-liguobin: Scala doesn't have `getOrDefault ` method. If you want to use it, you have to use Java HashMap.
```
import java.util.{HashMap => JHashMap}
val msp = new JHashMap()
```
or use `getOrElse` in scala.

src/main/resources/templates/kotlin-lang/kotlinClassGraphqlResponseProjection.ftl
jxnu-liguobin: please use this way:
```
    val projectionDepthOnFields2: MutableMap<String, Int> = mutableMapOf()
    projectionDepthOnFields2["key"] = value
```

jxnu-liguobin (changes requested):

plugins/sbt/graphql-java-codegen-sbt-plugin/src/main/scala/io/github/dreamylost/graphql/codegen/GraphQLCodegenPlugin.scala
```diff
@@ -114,6 +114,7 @@ class GraphQLCodegenPlugin(configuration: Configuration, private[codegen] val co
     generateToString := MappingConfigConstants.DEFAULT_TO_STRING,
     // parent interfaces configs:
     parentInterfaces := parentInterfacesConfig,
+    generateAllMethodInProjection := MappingConfigConstants.DEFAULT_GENERATE_ALL_METHOD,
```
jxnu-liguobin, line 117: This key must declare in `GraphQLCodegenKeys.scala `

jxnu-liguobin: e.g: `val generateAllMethodInProjection = settingKey[Boolean](you description for this option)`

src/main/resources/templates/scala-lang/scalaClassGraphqlResponseProjection.ftl
robbertnoordzij: Good improvement, do you know how to build these locally? I am getting an error that the 5.0.1-SNAPSHOT cannot be found

jxnu-liguobin (changes requested):

src/main/resources/templates/scala-lang/scalaClassGraphqlResponseProjection.ftl
jxnu-liguobin: for sbt
image

for gradle
image

src/main/resources/templates/scala-lang/scalaClassGraphqlResponseProjection.ftl
```diff
@@ -26,10 +29,12 @@ import java.util.Objects
 </#list>
 class ${className} extends GraphQLResponseProjection {
 
-<#if fields?has_content>
-    override def all$(): ${className} = all$(${responseProjectionMaxDepth})
+<#if fields?has_content && generateAllMethodInProjection>
+    var projectionDepthOnFields = new HashMap[String, Int]
```
jxnu-liguobin, line 33: please use `val`, because javamap is mutable struction.

robbertnoordzij: I changed it to val en the scala native hashmap with getOrElse

src/main/resources/templates/scala-lang/scalaClassGraphqlResponseProjection.ftl
robbertnoordzij: I changed it to the scala native HashMap with getOrElse

src/main/resources/templates/kotlin-lang/kotlinClassGraphqlResponseProjection.ftl
robbertnoordzij: Good improvement thank you! I changed them.

src/main/resources/templates/scala-lang/scalaClassGraphqlResponseProjection.ftl
```diff
@@ -26,16 +29,18 @@ import java.util.Objects
 </#list>
 class ${className} extends GraphQLResponseProjection {
 
-<#if fields?has_content>
-    override def all$(): ${className} = all$(${responseProjectionMaxDepth})
+<#if fields?has_content && generateAllMethodInProjection>
+    val projectionDepthOnFields = new HashMap[String, Int]
```
jxnu-liguobin, line 33: use `private final  lazy val projectionDepthOnFields = new HashMap[String, Int]`

src/main/resources/templates/kotlin-lang/kotlinClassGraphqlResponseProjection.ftl
```diff
@@ -26,10 +26,12 @@ import java.util.Objects
 </#list>
 open class ${className} : GraphQLResponseProjection() {
 
-<#if fields?has_content>
-    override fun `all$`(): ${className} = `all$`(${responseProjectionMaxDepth})
+<#if fields?has_content && generateAllMethodInProjection>
+    var projectionDepthOnFields: MutableMap<String, Int> = mutableMapOf()
```
jxnu-liguobin, line 30:
```
private val projectionDepthOnFields: MutableMap<String, Int> = mutableMapOf()
```

src/test/resources/expected-classes/kt/SearchResultItemResponseProjection.kt.txt
```diff
@@ -13,9 +13,11 @@ import java.util.Objects
 )
 open class SearchResultItemResponseProjection : GraphQLResponseProjection() {
 
-    override fun `all$`(): SearchResultItemResponseProjection = `all$`(3)
+    var projectionDepthOnFields: MutableMap<String, Int> = mutableMapOf()
```
jxnu-liguobin, line 16: Don't use `var`.

jxnu-liguobin (changes requested):

jxnu-liguobin:
It's basically OK, only a small optimization. 👍

src/main/resources/templates/kotlin-lang/kotlinClassGraphqlResponseProjection.ftl
robbertnoordzij: Nice suggestion, but I think it should be:

```kotlin
    private final val projectionDepthOnFields: MutableMap<String, Int> by mutableMapOf()
````

robbertnoordzij:
I saw Sonar is failing with a NullPointerException, is that related to the change? I cannot find the report for it.

src/main/resources/templates/scala-lang/scalaClassGraphqlResponseProjection.ftl
robbertnoordzij: Thanks for the suggestion!

kobylynskyi:
@robbertnoordzij GitHub does not share a secret for forks, so Sonar step is always failing for PRs created from forked repos... Sorry about that, but I can't find a solution to resolve this.

kobylynskyi (approved):
@robbertnoordzij looks good to me now. Good job! :+1:

jxnu-liguobin (approved):
I've tested the scala/kotlin client call, no problem.

kobylynskyi:
@robbertnoordzij thanks a lot for your contribution.
I will release this feature as part of the next release (most likely 5.1.0). Stay tuned ;)

Issue #660, opened 2021-04-19
OBJECT Directives not being mapped to Java annotations


TGNThump:
## Issue Description

OBJECT Directives are not being mapped to Java annotations

## Steps to Reproduce

```java
directive @Document(
    collection: String
) on OBJECT

type Test @Document(collection: "test") {
  name: String
}
```

## Expected Result

```java
@javax.annotation.processing.Generated(
    value = "com.kobylynskyi.graphql.codegen.GraphQLCodegen",
    date = "2021-04-19T15:35:57+0100"
)
@org.springframework.data.mongodb.core.mapping.Document("test")
public class Test implements java.io.Serializable {
...
```

## Actual Result

```java
@javax.annotation.processing.Generated(
    value = "com.kobylynskyi.graphql.codegen.GraphQLCodegen",
    date = "2021-04-19T15:35:57+0100"
)
public class Test implements java.io.Serializable {
...
```

## Your Environment and Setup

* **graphql-java-codegen version**: *5.0.0*
* **Build tool**: *Maven*
* **Mapping Config**:

```xml
...
 <directiveAnnotationsMapping>
    <Document>
        <annotation>org.springframework.data.mongodb.core.mapping.Document({{collection}})</annotation>
    </Document>
</directiveAnnotationsMapping>
...
```

robbertnoordzij:
I am able to reproduce it. In the annotation processing for a type, the directive is not added/parsed.

For now (I understand it is not ideal) you can add an customAnnotation to for each type:

```xml
                            <customAnnotationsMapping>
                                <Test>
                                    <annotation>
                                        @org.springframework.data.mongodb.core.mapping.Document("test")
                                    </annotation>
                                </Test>
                            </customAnnotationsMapping>
```

robbertnoordzij:
I think I was able to pin point the problem. However, I am not sure if this is done by accident or on purpose.

kobylynskyi:
Thanks @robbertnoordzij for the changes. Will be released as part of 5.1.0

PR #664, opened 2021-04-19
Release 5.1.0

kobylynskyi:
* Ability to disable generation of the all$ method in projections #645
* Generate jackson type id resolver annotation #637
* Add directive support for types #660 #665
* GraphQL serialization of java.util.Map #668 #669


jxnu-liguobin:
image
have a bug, NPE

jxnu-liguobin:
> > image
> > have a bug, NPE
> 
> Could the definition ever be null? I understand why sonar thinks it (because of the conditional) but will it ever happen?

Maybe because there is a check on the front, sonar thinks that the following also needs to be checked

kobylynskyi:
Agree with @jxnu-liguobin 
If the variable is null-checked, then all subsequent usages of this variable are considered nullable.
So we can either remove this null-check (if `extendedDefinition` is always non-null), or add null-check at all usages of `extendedDefinition`

robbertnoordzij:
> Agree with @jxnu-liguobin
> If the variable is null-checked, then all subsequent usages of this variable are considered nullable.
> So we can either remove this null-check (if `extendedDefinition` is always non-null), or add null-check at all usages of `extendedDefinition`

Not sure if someone was already fixing it, but I created an PR: https://github.com/kobylynskyi/graphql-java-codegen/pull/666

robbertnoordzij:
I noticed checkstyle does not run with gradle and not on the PR's. I see in this branch it found 2 issues: 

- Whitespace
- Order of the getAnnotations method

I created an PR to fix them both: https://github.com/kobylynskyi/graphql-java-codegen/pull/667.

Hope this helps.

Issue #673, opened 2021-04-28
EnumValues cannot be used directiveAnnotationsMapping.


michael-simons:
## Issue Description

I cannot use directiveAnnotationsMapping with directives spotting enum values.

Code generation will fail during issues with the `ValueMapper`.

## Steps to Reproduce

Given

```graphql
type Actor {
    name: String
    movies: [Movie] @relationship(type: "ACTED_IN", direction: OUT)
}
```

with a custom directive `@relationship`, using enum `OUT` on attribute `direction`

And the following config

```xml
<directiveAnnotationsMapping>
	<relationship>
		<f>@org.springframework.data.neo4j.core.schema.Relationship(type = {{type}}, direction = {{direction}})</f>
	</relationship>
</directiveAnnotationsMapping>
```

## Expected Result

I expect the following generated code:

```java
@org.springframework.data.neo4j.core.schema.Relationship(type = "ACTED_IN", direction = OUT)
private java.util.List<Movie> movies;
```

## Actual Result

Value Mapper fails:

java.lang.IllegalArgumentException: Unexpected Enum value for list type
    at com.kobylynskyi.graphql.codegen.mapper.ValueMapper.mapEnum (ValueMapper.java:132)
    at com.kobylynskyi.graphql.codegen.mapper.ValueMapper.map (ValueMapper.java:110)

When using a custom annotation mapping, the `GraphQLTypeMapper` will always call ```com.kobylynskyi.graphql.codegen.mapper.ValueMapper#map(com.kobylynskyi.graphql.codegen.model.MappingContext, graphql.language.Value<?>, graphql.language.Type<?>, java.lang.String)``` with an argument of `null` for the GraphQL type from 

`com.kobylynskyi.graphql.codegen.mapper.GraphQLTypeMapper#getAnnotationsForDirective` which will eventually trigger the illegal argument exception in `com.kobylynskyi.graphql.codegen.mapper.ValueMapper#mapEnum`.

## Your Environment and Setup

Maven,

```xml
<plugin>
    <groupId>io.github.kobylynskyi</groupId>
    <artifactId>graphql-codegen-maven-plugin</artifactId>
    <version>5.1.0</version>
    <executions>
	    <execution>
		    <goals>
			    <goal>generate</goal>
		    </goals>
		    <configuration>
			    <!-- all config options:
			    https://github.com/kobylynskyi/graphql-java-codegen/blob/master/docs/codegen-options.md
			    -->
[...]

PR #674, opened 2021-04-28
Don't fail with enums in directive parameters.

meistermeier:
---

### Description

Support mapping of enum values in directive parameters.
This does currently fail and I _think_ that this little bit extra for the `null`-handling can improve the situation.
I am pretty new to the whole GraphQL world and do not know if this makes sense at all.
Since I haven't found a similar test scenario for Kotlin, I skipped this part for this PR and just focused on Scala and Java.

Related to #673

---

Changes were made to:
- [x] Codegen library - Java
- [x] Codegen library - Scala

### DIFF
--- a/src/main/java/com/kobylynskyi/graphql/codegen/mapper/ValueMapper.java
+++ b/src/main/java/com/kobylynskyi/graphql/codegen/mapper/ValueMapper.java
@@ -129,6 +129,9 @@ public class ValueMapper {
         if (graphQLType instanceof NonNullType) {
             return mapEnum(mappingContext, value, ((NonNullType) graphQLType).getType());
         }
+        if (graphQLType == null) {
+            return value.getName();
+        }
         throw new IllegalArgumentException("Unexpected Enum value for list type");
     }
 
--- a/src/main/java/com/kobylynskyi/graphql/codegen/mapper/ValueMapper.java
+++ b/src/main/java/com/kobylynskyi/graphql/codegen/mapper/ValueMapper.java
@@ -121,6 +121,9 @@ public class ValueMapper {
     }
 
     private String mapEnum(MappingContext mappingContext, EnumValue value, Type<?> graphQLType) {
+        if (graphQLType == null) {
+            return value.getName();
+        }
         if (graphQLType instanceof TypeName) {
             String typeName = ((TypeName) graphQLType).getName();
             typeName = DataModelMapper.getModelClassNameWithPrefixAndSuffix(mappingContext, typeName);
@@ -128,9 +131,6 @@ public class ValueMapper {
         }
         if (graphQLType instanceof NonNullType) {
             return mapEnum(mappingContext, value, ((NonNullType) graphQLType).getType());
-        }
-        if (graphQLType == null) {
-            return value.getName();
         }
         throw new IllegalArgumentException("Unexpected Enum value for list type");
     }
--- a/src/test/resources/schemas/test.graphqls
+++ b/src/test/resources/schemas/test.graphqls
@@ -95,6 +95,12 @@ enum EventStatus {
     LOGGED
 }
 
+# type with directive using enum value
+type User {
+    name: String,
+    friends: [User] @relationship(type: "FRIEND_WITH", direction: OUT)
+}
+
 directive @auth (
     roles: [String] = ["viewer"],
     float: [Float] = [],
--- a/src/main/resources/templates/scala-lang/scalaClassGraphqlType.ftl
+++ b/src/main/resources/templates/scala-lang/scalaClassGraphqlType.ftl
@@ -15,22 +15,36 @@ import com.kobylynskyi.graphql.codegen.model.graphql.GraphQLRequestSerializer
 import java.util.Objects
 </#if>
 import scala.collection.JavaConverters._
+<#assign duplicateEnumImports = [] />
+<#assign enumImports = [] />
 <#if fields?has_content>
     <#if enumImportItSelfInScala?has_content>
         <#list fields as field>
             <#list enumImportItSelfInScala as enum>
                 <#if MapperUtil.isScalaCollection(field.type)>
                     <#if enum == MapperUtil.getGenericParameter(field.type)>
-import ${enum}._
+                        <#assign duplicateEnumImports = duplicateEnumImports + [enum] />
                     </#if>
                 <#else >
                     <#if enum == field.type>
-import ${enum}._
+                        <#assign duplicateEnumImports = duplicateEnumImports + [enum] />
                     </#if>
                 </#if>
             </#list>
         </#list>
     </#if>
+</#if>
+<#if duplicateEnumImports?has_content>
+    <#list duplicateEnumImports as duplicateEnumImport>
+        <#if !enumImports?seq_contains(duplicateEnumImport)>
+            <#assign enumImports = enumImports + [duplicateEnumImport]>
+        </#if>
+    </#list>
+</#if>
+<#if enumImports?has_content>
+    <#list enumImports as enumImport>
+import ${enumImport}._
+    </#list>
 </#if>
 <#assign duplicateParentInterfaces = [] />
 <#assign parentInterfaces = [] />
--- a/src/test/java/com/kobylynskyi/graphql/codegen/GraphQLCodegenAnnotationsTest.java
+++ b/src/test/java/com/kobylynskyi/graphql/codegen/GraphQLCodegenAnnotationsTest.java
@@ -210,6 +210,8 @@ class GraphQLCodegenAnnotationsTest {
                         "n={{n?toString}})"));
         directiveAnnotationsMapping.put("valid", singletonList("@javax.validation.Valid"));
         directiveAnnotationsMapping.put("customResolver", singletonList("@com.example.CustomAnnotation"));
+        directiveAnnotationsMapping.put("relationship",
+                singletonList("@com.example.Relationship(type = {{type}}, direction = {{direction}})"));
         mappingConfig.setDirectiveAnnotationsMapping(directiveAnnotationsMapping);
 
         new JavaGraphQLCodegen(singletonList("src/test/resources/schemas/test.graphqls"),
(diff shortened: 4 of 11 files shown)


src/test/resources/schemas/test.graphqls
```diff
@@ -95,6 +95,12 @@ enum EventStatus {
     LOGGED
 }
 
+# type with directive using enum value
+type User {
+    name: String,
+    friends: [User] @relationship(type: "FRIEND_WITH", direction: OUT)
```
jxnu-liguobin, line 101: Enumeration declaration of `OUT` should be provided, which is very important for Scala. It will have one more import statement. Enumeration not imported is not available. in scala.

Then, class will be:
```
import package.model._
import OUTENUM._
import scala.collection.JavaConverters._

/**
 * type with directive using enum value
 */
@javax.annotation.Generated(
    value = Array("com.kobylynskyi.graphql.codegen.GraphQLCodegen"),
    date = "2020-12-31T23:59:59-0500"
)
case class User(
    name: String,
    @com.example.Relationship(type = "FRIEND_WITH", direction = OUT)
    friends: scala.Seq[User]
) {

}
```

jxnu-liguobin (changes requested):

src/main/java/com/kobylynskyi/graphql/codegen/mapper/ValueMapper.java
```diff
@@ -129,6 +129,9 @@ private String mapEnum(MappingContext mappingContext, EnumValue value, Type<?> g
         if (graphQLType instanceof NonNullType) {
             return mapEnum(mappingContext, value, ((NonNullType) graphQLType).getType());
         }
+        if (graphQLType == null) {
```
kobylynskyi, line 132: I think that such null-checks should be done before `if (graphQLType instanceof NonNullType)` and `f (graphQLType instanceof TypeName)`, so can you please move this verification to the top of the method?
Thanks.

kobylynskyi (changes requested):

src/main/java/com/kobylynskyi/graphql/codegen/mapper/ValueMapper.java
jxnu-liguobin: 👍

meistermeier:
Thanks for your swift feedback. I wonder if it makes sense to support Scala in the scope of this PR. As far as I interpret the code around the `...DataModelMapper` the annotations are just `Strings` and would need to get elevated into a class on their own, or?
I don't mind doing this but this will probably blur the focus of the PR.

kobylynskyi:
@meistermeier, thanks for your contributions. I think that it would be good to make end-to-end changes in a single PR in order to keep track what was made and what's still pending (Java, Scala, Kotlin + Maven, Gradle, SBT)
Please let me know your thoughts.

jxnu-liguobin:
There is no problem with this PR change. I just want us to try our best to make the content of the test file tend to be correct for future reference.

meistermeier:
Was investigating the needed changes (for Scala) and found out that it is not only that the annotation needs to be sth. like an `AnnotationDefinition` but also it needs to get hydrated from the directive information. But I cannot see any suitable place in the code where I could do:
Create type definition -> create fields -> create annotations -> :no_entry: from annotation to directive meta-information to retrieve the type of the parameters :no_entry:
So I would suggest (and I can do this) to remove the Scala path and stick with a plain Java approach for now.

Unrelated but since I was on the Scala imports: I removed duplicates from the imports for Enums in the template. (https://github.com/kobylynskyi/graphql-java-codegen/pull/674/commits/8d630278eb2676c8f0e449bd5a11f1084e2af318)

jxnu-liguobin:
> Unrelated but since I was on the Scala imports: I removed duplicates from the imports for Enums in the template. ([8d63027](https://github.com/kobylynskyi/graphql-java-codegen/commit/8d630278eb2676c8f0e449bd5a11f1084e2af318))

Thanks, the implementation looks good to me, just like the existing one. However, we may need to use `HashSet` to transfer to the template later to avoid the template being too complex. But this PR does not need to considere it.

kobylynskyi (approved):

jxnu-liguobin:
I just think PR lacks some test case. Although logically there is no problem. If you are in a hurry, you can merge.

jxnu-liguobin:
> There is right now the somehow wrong `User.scala` test in place where I check for the right annotation.
> In there is the enum that does not declare the import for the enum due to the lack of infrastructure as I mentioned [#674 (comment)](https://github.com/kobylynskyi/graphql-java-codegen/pull/674#issuecomment-829230812)

Because enumeration `Out` must also be declared in `src/test/resources/schemas/test.graphqls`, then document parser can get it for scala and has a import.

meistermeier:
Yes, I understand this but the problem is with
```
type User {
    name: String,
    friends: [User] @relationship(type: "FRIEND_WITH", direction: OUT)
}

enum Direction {
    IN,
    OUT
}
```
the import will not get triggered in the Scala world. Nowhere to be precise but Java does not complain about this.

Even adding a 
```
directive @relationship (
    type: String,
    direction: Direction
) on FIELD_DEFINITION
```
won't help because the meta-information are not telling me anything about types in the annotation.
This would need a lot of refactoring -if even possible- to get those information.

Thus my initial question was if I should remove the assumptions in the Scala test code completely.

jxnu-liguobin:
The problem here may be that `ftl` does not contains enumeration in the fields, so it is a bug.

Because enumeration in annotation is not in `fields`.
```ftl
<#if fields?has_content>
    <#if enumImportItSelfInScala?has_content>
        <#list fields as field>
            <#list enumImportItSelfInScala as enum>
                <#if MapperUtil.isScalaCollection(field.type)>
                    <#if enum == MapperUtil.getGenericParameter(field.type)>
import ${enum}._
                    </#if>
                <#else >
                    <#if enum == field.type>
import ${enum}._
                    </#if>
                </#if>
            </#list>
        </#list>
    </#if>
</#if>
```

And it's not easy to handle. If processing is forced to be added to the template, the parameter of the annotation should use `Direction.OUT`, which makes the template more complicated. In view of this, the PR no longer needs to be related to the problem, I will create a new issue to track it.

jxnu-liguobin (approved):

jxnu-liguobin:
Another way is to generate the following annotation:
```
@com.example.Relationship(type = "FRIEND_WITH", direction = DIRECTION.OUT)
```

I'm not sure whether graphql supports `DIRECTION.OUT`, but for Scala, if can use `DIRECTION.OUT` that the template will be a little simpler. Otherwise we need to judge and import enumeration.

A new issue trace has been created for it. It may be resolved later.

kobylynskyi:
@jxnu-liguobin @meistermeier is this ready to be merged to develop and eventually be released as part of the next release (5.3.0)?

jxnu-liguobin:
> @jxnu-liguobin @meistermeier is this ready to be merged to develop and eventually be released as part of the next release (5.3.0)?

Yes

jxnu-liguobin:
@tejas-sastry Please show the failed code and schema.

tejas-sastry:
schema:
```
directive @NotNull(message : String, groups: ValidationGroup) on INPUT_FIELD_DEFINITION

input TrustAccountInput
{
  accountId: String! @NotNull(message: "test", groups: CAMS)
}

```
Plugin config:
```
<plugin>
                <groupId>io.github.kobylynskyi</groupId>
                <artifactId>graphql-codegen-maven-plugin</artifactId>
                <version>${graphqlcodegen-maven-plugin.version}</version>
                <executions>
                    <execution>
                        <goals>
                            <goal>generate</goal>
                        </goals>
                    </execution>
                </executions>
                <configuration>
                    <graphqlSchemas>
                        <rootDir>${project.basedir}/src/main/resources/graphql/</rootDir>
                    </graphqlSchemas>
                    <generatedLanguage>KOTLIN</generatedLanguage>
                    <generateEqualsAndHashCode>true</generateEqualsAndHashCode>
                    <outputDir>${project.basedir}/target/generated</outputDir>
                    <addGeneratedAnnotation>false</addGeneratedAnnotation>
                    <generateClient>true</generateClient>
                    <packageName>com.intuit.identity.manage.model</packageName>
                    <modelPackageName>com.intuit.identity.manage.model.graphql</modelPackageName>
                    <apiPackageName>com.intuit.identity.manage.model.querybuilders</apiPackageName>
                    <initializeNullableTypes>true</initializeNullableTypes>
                    <directiveAnnotationsMapping>
                        <ValidAccountId>
                            <f>
                                @com.intuit.identity.manage.account.model.constraints.ValidAccountId
                            </f>
                        </ValidAccountId>
                        <NotBlank>
                            <f>
                                @javax.validation.constraints.NotBlank
                            </f>
                        </NotBlank>
                        <NotNull>
                            <f>
                                @javax.validation.constraints.NotNull(message = {{message}}, groups = {{groups}})
                            </f>
                        </NotNull>
                        <NotNullCams>
                            <f>
                                @javax.validation.constraints.NotNull(message = {{message}}, groups = [com.intuit.identity.manage.enum.CamsGroup::class])
                            </f>
                        </NotNullCams>
                        <Valid>
                            <f>
                                @javax.validation.Valid
                            </f>
                        </Valid>
                        <Validated>
                            <f>
                                @org.springframework.validation.annotation.Validated
                            </f>
                        </Validated>
                    </directiveAnnotationsMapping>
                    <customTypesMapping>
                        <Date>java.time.LocalDate</Date>
                        <DateTime>java.time.LocalDateTime</DateTime>
                        <CAMS>com.intuit.identity.manage.enum.CamsGroup::class</CAMS>
                    </customTypesMapping>
                    <useObjectMapperForRequestSerialization>
                        Date,DateTime
                    </useObjectMapperForRequestSerialization>
                </configuration>
            </plugin>
```

Generated class:
```

import com.kobylynskyi.graphql.codegen.model.graphql.GraphQLRequestSerializer
import java.util.StringJoiner

/**
 * Input for providing trust account in different mutation variations
 */
data class TrustAccountInput(
    @field:javax.validation.constraints.NotNull(message = "test", groups = CAMS)
    var accountId: String,
    var trustGrant: TrustGrantInput,
    var primaryProfile: PrimaryProfileInput? = null
) {...}
```
Error: `Unresolved reference: CAMS`

My expectation is that the `custom type mapping` replaces the enum.

jxnu-liguobin:
@tejas-sastry Thank you for your report. This is indeed a problem. In kotlin, customTypesMapping only handles primitive types. I have opened a new issue for it  andthis is a legacy issue left over from the first support for kotlin

Issue #681, opened 2021-05-06
Ability to provide a custom operation name in GraphQL request


robbertnoordzij:
**Is your feature request related to a problem? Please describe.**
Currently we use the same generated code in different clients. However, we would like to distinguish between the sources of the request. An option would be to use the operation name as specified on https://graphql.org/learn/queries/#operation-name. At the moment this field seems to be static and final in the Request class. And the operation is also used as the actual field of the query. 

E.g. the currently the generated `HeroesQueryRequest` has the `OPERATION_NAME` set (final) to `hero`. In the `GraphQLRequestSerializer` this operation name is being used in the operation name and query part of the request.

Is this intentional?

**Describe the solution you'd like**
In this example, I would like to be able to set the operation name to `HeroNameAndFriends` and stil request the field hero from the query.

**Describe alternatives you've considered**
I have tried to use the alias. However, this only directs the output of the query to another field and this won't change the operation name as extracted by our graphql server.

kobylynskyi:
@robbertnoordzij good point. This could be enhanced by providing a custom query/mutation name.
The changes should be done in `GraphQLRequest` by creating another constructor(s) with an additional `String` argument.

robbertnoordzij:
@kobylynskyi not to put pressure on to the project or to push anything, do you have an estimated release date?

kobylynskyi:
@robbertnoordzij, sorry for the delay. I've released 5.2.0 just now.

robbertnoordzij:
No problem! Thank you!

PR #682, opened 2021-05-06
Ability to add custom operation name for request

robbertnoordzij:
---

### Description

Based on the suggestion by @kobylynskyi, I added the ability to override the operation name in queries. While to comment states that a multi request does not have an operation name it is actually possible based on the documentation.

Am I missing something?

Related to #681 

---

Changes were made to:
- [ ] Codegen library - Java
- [ ] Codegen library - Scala
- [ ] Maven plugin
- [ ] Gradle plugin
- [ ] SBT plugin


src/main/java/com/kobylynskyi/graphql/codegen/model/graphql/GraphQLRequest.java
```diff
@@ -7,14 +7,24 @@
...
 
     public GraphQLRequest(GraphQLOperationRequest request, GraphQLResponseProjection responseProjection) {
+        this(request, responseProjection, null);
+    }
+
+    public GraphQLRequest(
+        GraphQLOperationRequest request,
+        GraphQLResponseProjection responseProjection,
+        String customOperationName
+    ) {
```
kobylynskyi, line 24: IMHO the order of parameters should correspond to the order entities in GraphQL request ("natural" order), meaning that when you construct the query, you first define the name of the operation, then request, and then response projection. So 2 points I would like to change here:
1. Make `String operationName` as a first parameter in constructor(s).
2. Create another constructor `(String operationName, GraphQLOperationRequest request)`.

Open to the descission.

src/main/java/com/kobylynskyi/graphql/codegen/model/graphql/GraphQLRequests.java
```diff
@@ -10,8 +10,14 @@
 public class GraphQLRequests {
 
     private final List<GraphQLRequest> requests = new ArrayList<>();
+    private final String customOperationName;
```
kobylynskyi, line 13: Please change the order of the fields to match the order of constructor parameters.

kobylynskyi (changes requested):

src/main/java/com/kobylynskyi/graphql/codegen/model/graphql/GraphQLRequest.java
robbertnoordzij: Good points. I will make the improvements tomorrow or the day after.

src/main/java/com/kobylynskyi/graphql/codegen/model/graphql/GraphQLRequests.java
```diff
@@ -10,8 +10,14 @@
 public class GraphQLRequests {
 
     private final List<GraphQLRequest> requests = new ArrayList<>();
+    private final String operationName;
```
kobylynskyi, line 13:
```suggestion
    private final String operationName;
    private final List<GraphQLRequest> requests = new ArrayList<>();
```

kobylynskyi: (to be in sync with the order of arguments in constructor)

src/main/java/com/kobylynskyi/graphql/codegen/model/graphql/GraphQLRequest.java
```diff
@@ -5,14 +5,28 @@
...
 
     public GraphQLRequest(GraphQLOperationRequest request, GraphQLResponseProjection responseProjection) {
+        this(null, request, responseProjection);
+    }
+
+    public GraphQLRequest(
+        String operationName,
+        GraphQLOperationRequest request,
+        GraphQLResponseProjection responseProjection
+    ) {
```
kobylynskyi, line 28:
```suggestion
    public GraphQLRequest(String operationName, GraphQLOperationRequest request,
        GraphQLResponseProjection responseProjection) {
```

kobylynskyi: (as per the formating rules located in ./config/checkstyle/graphql-codegen-check-style.xml)

kobylynskyi (changes requested):

kobylynskyi (approved):
@robbertnoordzij Good job. Thanks for the contribution.

Issue #810, opened 2021-09-14
Add serialVersionUID to generated classes that implement Serializable


rolevinks:
We use the maven-compiler-plugin in our project, and this gives us warnings on generated classes that implement Serializable.

`serializable class [...] has no definition of serialVersionUID`

There is a simple, solution, to add this to the template:

```
    @java.io.Serial
    private static final long serialVersionUID = 1L;
```

Can this be implemented? I would do it myself, but when I check out and build the project, a lot of tests fail because of an UnableToDeleteDirectoryException.

```
Unable to delete directory
com.kobylynskyi.graphql.codegen.model.exception.UnableToDeleteDirectoryException: Unable to delete directory

Caused by: java.nio.file.FileSystemException: build\generated\com\kobylynskyi\graphql\test1\CreateEventMutationResolver.java: The process cannot access the file because it is being used by another process
	at java.base/sun.nio.fs.WindowsException.translateToIOException(WindowsException.java:92)
	at java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:103)
	at java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:108)
	at java.base/sun.nio.fs.WindowsFileSystemProvider.implDelete(WindowsFileSystemProvider.java:274)
	at java.base/sun.nio.fs.AbstractFileSystemProvider.delete(AbstractFileSystemProvider.java:105)
	at java.base/java.nio.file.Files.delete(Files.java:1146)
	at com.kobylynskyi.graphql.codegen.utils.Utils.deleteDir(Utils.java:161)
	... 101 more 
```

rolevinks:
Is resolved with https://github.com/kobylynskyi/graphql-java-codegen/pull/811

PR #811, opened 2021-09-15
Add serialVersionUID to generated classes that implement Serializable

rolevinks:
---

This pull request adds a serialVersionUID to the java template _type.ftl_. This prevent warnings from for example the maven-compiler-plugin, that are raised when a class that implements Serializable does not has a serialVersionUID field.

Related to #810

---

Changes were made to:
- [x] Codegen library - Java
- [ ] Codegen library - Scala
- [ ] Maven plugin
- [ ] Gradle plugin
- [ ] SBT plugin

### DIFF
--- a/src/main/resources/templates/java-lang/type.ftl
+++ b/src/main/resources/templates/java-lang/type.ftl
@@ -36,6 +36,9 @@ import java.util.StringJoiner;
 </#list>
 public class ${className} implements java.io.Serializable<#if implements?has_content><#list implements as interface>, ${interface}<#if interface_has_next></#if></#list></#if> {
 
+    @java.io.Serial
+    private static final long serialVersionUID = 1L;
+
 <#if fields?has_content>
     <#list fields as field>
         <#if field.deprecated?has_content>
--- a/src/main/resources/templates/java-lang/type.ftl
+++ b/src/main/resources/templates/java-lang/type.ftl
@@ -36,7 +36,6 @@ import java.util.StringJoiner;
 </#list>
 public class ${className} implements java.io.Serializable<#if implements?has_content><#list implements as interface>, ${interface}<#if interface_has_next></#if></#list></#if> {
 
-    @java.io.Serial
     private static final long serialVersionUID = 1L;
 
 <#if fields?has_content>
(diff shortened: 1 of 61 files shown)


kobylynskyi:
@rolevinks please change target branch to **develop**

rolevinks:
@kobylynskyi Yes, sorry! Target branch is now develop

src/main/resources/templates/java-lang/type.ftl
```diff
@@ -36,6 +36,9 @@ import java.util.StringJoiner;
 </#list>
 public class ${className} implements java.io.Serializable<#if implements?has_content><#list implements as interface>, ${interface}<#if interface_has_next></#if></#list></#if> {
 
+    @java.io.Serial
+    private static final long serialVersionUID = 1L;
```
jxnu-liguobin, line 40: welcome to create pr.
I only have one questions. 
1. The library is based on 1.8, and the annotation comes from Java 11.

So It is not appropriate to introduce this annotation directly. With configuration, placeholders can be more generic for compatibility with Java 1.8 and other languages.

jxnu-liguobin (changes requested):

src/main/resources/templates/java-lang/type.ftl
jxnu-liguobin: https://docs.oracle.com/en/java/javase/14/docs/api/java.base/java/io/Serial.html

Add: officially introduced in the  Java 14.

rolevinks: Ah yes, I removed the annotation. It is not really needed anyway

rolevinks: Hi, 
Is there any update on this issue?

jxnu-liguobin: Maybe we can check the JDK version in the template. 

`<#assign MapperUtil=statics["com.kobylynskyi.graphql.codegen.java.Utils"]>`

and `<# if checkJdk />` ? @kobylynskyi

kobylynskyi (approved):

src/main/resources/templates/java-lang/type.ftl
kobylynskyi: I don't think we need that annotation anyway.
Sorry for a long delay in reply!

PR #866, opened 2021-11-15
When using GraphQL federation, external parties can extend types and …

aldib:
---

### Description

When using GraphQL federation, external parties can extend types and inputs from the initial schema. Consequently, the server might receive queries and mutations that contains more fields than the ones originally generated.

This changes introduces two new options:
| Option                                                | Data Type                                                             | Default value                                          | Description |
| :---------------------------------------------------: | :-------------------------------------------------------------------: | :----------------------------------------------------: | ----------- |
| `supportUnknownFields`                                | Boolean                                                               | false                                                  | Specifies whether api classes should support unknown fields during serialization or deserialization. If `true`, classes will include a property of type [`java.util.Map<String,Object>`](https://docs.oracle.com/javase/8/docs/api/index.html?java/util/Map.html) that will store unknown fields.|
| `unknownFieldsPropertyName`                           | String                                                                | userDefinedFields                                      | Specifies the name of the property to be included in api classes to support unknown fields during serialization or deserialization|

When `supportUnknownFields=true` a new catch-all field annotated with the [com.fasterxml.jackson.annotation.JsonAnySetter](https://fasterxml.github.io/jackson-annotations/javadoc/2.12/com/fasterxml/jackson/annotation/JsonAnySetter.html) and [com.fasterxml.jackson.annotation.JsonAnyGetter](https://fasterxml.github.io/jackson-annotations/javadoc/2.12/com/fasterxml/jackson/annotation/JsonAnyGetter.html) is added to generated code to serialize and deserialize unknown fields using a map.

Please let me know if you agree with the changes and/or if I missed anything.

Thank you in advance

---

Changes were made to:
- [x] Codegen library - Java
- [x] Codegen library - Scala
- [x] Maven plugin
- [x] Gradle plugin
- [x] SBT plugin


plugins/gradle/example-server/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,5 @@
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.1-bin.zip
+zipStoreBase=GRADLE_USER_HOME
+zipStorePath=wrapper/dists
```
kobylynskyi, line 5: @aldib could you please remove gradle wrapper files from example-server?
You can build any subdirectory using:
`./gradlew -p plugins/gradle/example-server clean build --warning-mode all`

src/main/java/com/kobylynskyi/graphql/codegen/mapper/InputDefinitionToDataModelMapper.java
```diff
@@ -56,8 +63,20 @@ public InputDefinitionToDataModelMapper(MapperFactory mapperFactory,
...
+        List<ParameterDefinition> fields = inputValueDefinitionToParameterMapper
+                .map(mappingContext, definition.getValueDefinitions(), definition.getName());
+        if (mappingContext.isSupportUnknownFields()){
+            ParameterDefinition unknownFields = new ParameterDefinition();
+            unknownFields.setName(mappingContext.getUnknownFieldsPropertyName());
+            unknownFields.setOriginalName(mappingContext.getUnknownFieldsPropertyName());
+            unknownFields.setType("java.util.Map<String,Object>");
+            unknownFields.setAnnotations(Arrays.asList(
+                    JsonAnySetter.class.getName(),
+                    JsonAnyGetter.class.getName()
```
kobylynskyi, line 75: Shouldn't this also be configurable?

src/main/java/com/kobylynskyi/graphql/codegen/mapper/TypeDefinitionToDataModelMapper.java
```diff
@@ -130,6 +115,21 @@ private static ParameterDefinition merge(ParameterDefinition typeDef, ParameterD
...
+
+        if (mappingContext.isSupportUnknownFields() ){
+            ParameterDefinition unknownFields = new ParameterDefinition();
+            unknownFields.setName(mappingContext.getUnknownFieldsPropertyName());
+            unknownFields.setOriginalName(mappingContext.getUnknownFieldsPropertyName());
+            unknownFields.setType("java.util.Map<String,Object>");
+            unknownFields.setAnnotations(Arrays.asList(
+                    JsonAnySetter.class.getName(),
+                    JsonAnyGetter.class.getName()
+            ));
```
kobylynskyi, line 129: I see you have exact same instantiation of `ParameterDefinition` in different pleases. Please extract it.

src/main/java/com/kobylynskyi/graphql/codegen/model/MappingConfig.java
```diff
@@ -1,5 +1,9 @@
 package com.kobylynskyi.graphql.codegen.model;
 
+import com.fasterxml.jackson.annotation.JsonAnyGetter;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Target;
```
kobylynskyi, line 6: I think this imports are redundant

kobylynskyi (changes requested):
Overall the idea and implementation look ok to me. Please address the requested changes and also `check-code-style` issues.
Thank you!

plugins/gradle/example-server/gradle/wrapper/gradle-wrapper.properties
aldib: Removed. Thank you

src/main/java/com/kobylynskyi/graphql/codegen/mapper/InputDefinitionToDataModelMapper.java
aldib: I am not sure what you mean for configurable. Both annotations are required to handle unknown properties, and they only take a `boolean enabled() default true;` argument. What type of configuration are you referring to?

src/main/java/com/kobylynskyi/graphql/codegen/mapper/TypeDefinitionToDataModelMapper.java
aldib: The logic is now externalized in `com.kobylynskyi.graphql.codegen.mapper.UnknownFieldsSupport` and used in both  `com.kobylynskyi.graphql.codegen.mapper.InputDefinitionToDataModelMapper` and  `com.kobylynskyi.graphql.codegen.mapper.TypeDefinitionToDataModelMapper`

src/main/java/com/kobylynskyi/graphql/codegen/model/MappingConfig.java
aldib: Removed. Thank you

src/main/java/com/kobylynskyi/graphql/codegen/mapper/InputDefinitionToDataModelMapper.java
aldib: I had a look at Jackson to see if there are other ways of configuring support for unknown properties, and I think the two annotations above is the only option. I'll resolve this conversation on the assumption that the code is ok as is, but happy to change it if needed.

kobylynskyi: @aldib I was taking about ability for users to choose which annotation to use. What if someone prefers other ser/deser provider like Gson?

aldib: Hi,

I agree with your idea, but the dependency on Jackson is already hard coded in a number of places. For example:
* https://github.com/kobylynskyi/graphql-java-codegen/blob/97fef9681501a883e49ab4a482f4e5d01e982230/src/main/java/com/kobylynskyi/graphql/codegen/parser/GraphQLDocumentParser.java#L3
* https://github.com/kobylynskyi/graphql-java-codegen/blob/97fef9681501a883e49ab4a482f4e5d01e982230/src/main/java/com/kobylynskyi/graphql/codegen/supplier/MergeableMappingConfigSupplier.java#L3

If you see it as a requirement, I can add an additional flag to specify the serialization library (e.g. jackson vs gson), and modify the factory method accordingly. However, I think that it might introduce inconsistency unless the same principle is applied everywhere else.

If you are happy with the code as is, I would appreciate if you could resolve this conversation and authorize the next steps of the merge process.

Thank you.

src/main/java/com/kobylynskyi/graphql/codegen/mapper/TypeDefinitionToDataModelMapper.java
```diff
@@ -132,6 +139,12 @@ private static ParameterDefinition merge(ParameterDefinition typeDef, ParameterD
                 .flatMap(Collection::stream)
                 .forEach(paramDef -> allParameters
                         .merge(paramDef.getName(), paramDef, TypeDefinitionToDataModelMapper::merge));
+
+
+        createUnknownFields(mappingContext).ifPresent(
+                unknownFields->allParameters.put(mappingContext.getUnknownFieldsPropertyName(),unknownFields)
```
jxnu-liguobin, line 145: space？

src/main/java/com/kobylynskyi/graphql/codegen/mapper/UnknownFieldsSupport.java
```diff
@@ -0,0 +1,42 @@
...
+     * @param mappingContext The context of the mapping process.
+     * @return If {@link MappingContext#isSupportUnknownFields()} is true, it returns a monad containing
+     *         the instance of {@link ParameterDefinition}. {@link Optional#empty()} otherwise.
+     */
+    default Optional<ParameterDefinition> createUnknownFields(MappingContext mappingContext){
+        if (mappingContext.isSupportUnknownFields()){
+            ParameterDefinition unknownFields = new ParameterDefinition();
+            unknownFields.setName(mappingContext.getUnknownFieldsPropertyName());
+            unknownFields.setOriginalName(mappingContext.getUnknownFieldsPropertyName());
+            unknownFields.setType("java.util.Map<String,Object>");
```
jxnu-liguobin, line 33: `java.util.Map<String, Object>`

src/main/java/com/kobylynskyi/graphql/codegen/mapper/TypeDefinitionToDataModelMapper.java
aldib: Fixed. I finally managed to install the custom check-style custom configuration. I kept getting the _Property 'scope' does not exist_ exception before downgrading Intellij

src/main/java/com/kobylynskyi/graphql/codegen/mapper/UnknownFieldsSupport.java
aldib: Fixed

plugins/sbt/graphql-java-codegen-sbt-plugin/.bsp/sbt.json
```diff
@@ -0,0 +1 @@
+{"name":"sbt","version":"1.4.7","bspVersion":"2.0.0-M5","languages":["scala"],"argv":["/usr/lib/jvm/java-1.8.0-openjdk-1.8.0.312.b07-1.el8_4.x86_64/jre/bin/java","-Xms100m","-Xmx100m","-classpath","/home/dibellaa/.local/share/JetBrains/Toolbox/apps/IDEA-U/ch-0/212.5457.46.plugins/Scala/launcher/sbt-launch.jar","xsbt.boot.Boot","-bsp"]}
```
jxnu-liguobin, line 1: This file should be ignored

aldib: I am not familiar with Scala and there are no instructions to build the SBT plugin.  Should I remove the entire .bsp folder or only that file?

aldib: Removed. Are there any instructions on how to build the SBT plugin? I am not familiar with Scala and struggling to commit the correct changes

src/main/java/com/kobylynskyi/graphql/codegen/mapper/InputDefinitionToDataModelMapper.java
aldib: @kobylynskyi  the build is failing because of `No sonar.login or SONAR_TOKEN env variable was set`. Can you please advise on how to proceed?

kobylynskyi:
@aldib I've addressed some of the changes in your branch, please review my changes and let me know if you have any questions.
Also, I see that some of your unit tests are failing. Please address those and we will be good to merge this PR.
Thanks!
And sorry for not responding for so long - I got pulled to other stuff.

aldib:
@kobylynskyi  I am not sure what to do about the failed sonar job. Could you please advise on how to solve the problem? Thanks

kobylynskyi (approved):

Issue #868, opened 2021-11-17
Support "sealed" keyword on interfaces and unions in Kotlin


kbrooks:
**Is your feature request related to a problem? Please describe.**
Add `sealed` keyword` to generated interfaces and unions in order to enable exhaustive type-checking in written code. For example
```
sealed interface Animal {}
data class Cat(val foo: String): Animal  
data class Dog(val bar: String): Animal  

fun cases(obj: Animal) {                                
    return  when (obj) {
        is Cat -> "foo"
        is Dog -> "bar"; // Removing this line will fail compilation 
    }   
}
```
The Kotlin language will default to exhaustive `when` _statements_ (expressions are already exhaustive) in version 1.7, which will make it easier to take advantage of this feature. https://kotlinlang.org/docs/whatsnew1530.html#exhaustive-when-statements-for-sealed-and-boolean-subjects

**Describe the solution you'd like**
A `generateSealedInterfaces` build parameter that if true will add the `sealed` keyword to generated interfaces in Kotlin

**Describe alternatives you've considered**
n/a

kbrooks:
I noticed Scala has a sealed class option as well, but it looks like you have to make sure the classes are in the same file as the interface which seems like a non-starter

jxnu-liguobin:
> I noticed Scala has a sealed class option as well, but it looks like you have to make sure the classes are in the same file as the interface which seems like a non-starter

Yes, so I don't think Scala applies.

kbrooks:
It looks like Java actually added support for this feature in v15. I can add this as well https://openjdk.java.net/jeps/409

kbrooks:
Since Java's sealed interfaces have a more complex syntax that requires listing the implementing classes in the interface definition, it is significantly harder to add support; we can track it in another issue

PR #869, opened 2021-11-17
[Kotlin] Ability to generate sealed interfaces

kbrooks:
---

### Description

Adds `sealed` keyword to generated interfaces and unions in order to enable exhaustive type-checking in written code. Only works in and in Kotlin, even though Scala has support for sealed interfaces, because Scala has the restriction that classes inheriting a sealed interface are in the same file as that interface. Java has support for sealed interfaces but requires a `permits A, B, ...` syntax that makes it more difficult to support since we would need to parse the entire schema for all values that implement a given interface.

Related to #868 

---

Changes were made to:
- [x] Codegen library - Java
- [ ] Codegen library - Scala
- [x] Maven plugin
- [x] Gradle plugin
- [ ] SBT plugin


kbrooks:
Sonar is failing for mysterious reasons

docs/codegen-options.md
```diff
@@ -57,6 +57,7 @@ See [DirectiveAnnotationsMapping](#option-directiveannotationsmapping)* |
...
 | `generateModelOpenClasses`                            | Boolean                                                               | false                                                  | The class type of the generated model. If true, generate normal classes, else generate data classes. It only support in kotlin(```data class```) and scala(```case class```). Maybe we will consider to support Java ```record``` in the future.|
 | `initializeNullableTypes`                             | Boolean                                                               | false                                                  | Adds a default null value to nullable arguments. Only supported in Kotlin.
+| `generateSealedInterfaces`                            | Boolean                                                               | false                                                  | This applies to generated interfaces on unions and interfaces in Kotlin. If true, generate sealed interfaces, else generate normal ones. It is only supported in kotlin. |
```
kbrooks, line 60:
```suggestion
| `generateSealedInterfaces`                            | Boolean                                                               | false                                                  | This applies to generated interfaces on unions and interfaces. If true, generate sealed interfaces, else generate normal ones. It is only supported in Kotlin. |
```

plugins/sbt/graphql-java-codegen-sbt-plugin/src/main/scala/io/github/dreamylost/graphql/codegen/GraphQLCodegenKeys.scala
```diff
@@ -124,6 +124,8 @@ trait GraphQLCodegenKeys {
 
   val generateModelOpenClasses = settingKey[Boolean]("The class type of the generated model. If true, generate normal classes, else generate case class.")
 
+  val generateSealedInterfaces = settingKey[Boolean]("If true, generate sealed interfaces for GraphQL unions and interface, else generate normal interfaces.")
```
kbrooks, line 127:
```suggestion
  val generateSealedInterfaces = settingKey[Boolean]("If true, generate sealed interfaces for GraphQL unions and interfaces, else generate normal interfaces.")
```

kbrooks:
java interface is missing "permits"

kobylynskyi (approved):
Looks good!

Issue #859, opened 2021-10-21
Parameterized field getters in type interface are missing parameters


UselessPickles:
## Issue Description

I would like to generate a type as an interface that has parameterized fields, without using separate field resolvers for the parameterized fields. Since the type is being generated as an interface rather than a concrete class, I would expect the getter methods for the parameterized fields to be properly parameterized. However, there are no such parameters on the getters for the parameterized fields, and the generated interface is invalid/unusable. 

## Steps to Reproduce

Excerpt of relevant graphql schema:
```
"used with typesAsInterfaces config"
directive @asInterface on OBJECT

"used with fieldsWithResolvers config"
directive @noResolver on FIELD_DEFINITION

type Foo @asInterface {
  simpleField: String!
  parameterizedField(count: Int!): String! @noResolver
}
```

## Expected Result

Expected generated interface:
```
public interface Foo {
    @javax.validation.constraints.NotNull
    String getSimpleField();

    @javax.validation.constraints.NotNull
    String getParameterizedField(int count);
}
```

## Actual Result

Generated interface:
```
public interface Foo {
    @javax.validation.constraints.NotNull
    String getSimpleField();

    // missing parameter here
    @javax.validation.constraints.NotNull
    String getParameterizedField();
}
```

This interface is unusable/broken. In addition to being unable to implement the method in terms of its parameters, this is not even valid for graphql kickstart:

> Failed to instantiate [graphql.kickstart.tools.SchemaParser]: Factory method 'schemaParser' threw exception; nested exception is graphql.kickstart.tools.resolver.FieldResolverError: No method found as defined in schema <unknown>:216 with any of the following signatures (with or without one of [interface graphql.schema.DataFetchingEnvironment] as the last argument), in priority order:
> 
>   com.workforcesoftware.timesheetservice.generated.graphql.model.Foo.parameterizedField(~count)
>   com.workforcesoftware.timesheetservice.generated.graphql.model.Foo.getParameterizedField(~count)

## Your Environment and Setup

* **graphql-java-codegen version**: *5.2.0*
* **Build tool**: *Gradle*
* **Config**:
Excerpt of `graphqlCodegen` from build.gradle
```
  generateApis = true
  apiInterfaceStrategy = "INTERFACE_PER_OPERATION"
  apiRootInterfaceStrategy = "DO_NOT_GENERATE"
  apiReturnType = "graphql.execution.DataFetcherResult"
[...]

PR #913, opened 2021-12-03
Parameterized field getters in type interface are missing parameters #859

kobylynskyi:
### Description

Related to #859

---

Changes were made to:
- [X] Codegen library - Java
- [ ] Codegen library - Scala


kobylynskyi:
Hi @jxnu-liguobin could you please have a look if same changes should be done in scala/kotlin .ftl files? Thank you!

jxnu-liguobin:
> Hi @jxnu-liguobin could you please have a look if same changes should be done in scala/kotlin .ftl files? Thank you!

Yes, it need do some work, but since kotlin and scala have treated the type interface as a field, this is a big change to scala and kotlin and must be changed to `method`

Now
```scala
trait ProfileOwner {

    val anyPinnableItems: Boolean

    val email: String
}
```

Chage to 
```scala
trait ProfileOwner {

    def anyPinnableItems(a: Int): Boolean

    def email(a: Int): String
}
```

So, we can create a new issue and follow the problem until there is enough value to do. (using method is not friendly in most cases)

PR #958, opened 2022-02-09
Add annotations to JavaBean getters

ludovicc:
### Description

Add nonnull and other annotations to the getter method in addition to the field declaration. Some Java Bean introspection tools can only use annotations declared on the getter/setter methods, not on the private fields, so we need to repeat the annotation that was placed on the field on the getter as well to get full introspection with annotations.

---

Changes were made to:
- [x] Codegen library - Java
- [ ] Codegen library - Scala
- [ ] Maven plugin
- [ ] Gradle plugin
- [ ] SBT plugin

### DIFF
--- a/src/main/resources/templates/java-lang/type.ftl
+++ b/src/main/resources/templates/java-lang/type.ftl
@@ -73,6 +73,9 @@ public class ${className} implements java.io.Serializable<#if implements?has_con
         <#if field.deprecated?has_content>
     @${field.deprecated.annotation}
         </#if>
+        <#list field.annotations as annotation>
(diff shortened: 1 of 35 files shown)


kobylynskyi:
@ludovicc looks good. but could you please change the target branch to **develop**?

kobylynskyi:
@ludovicc can you please fix the build?

JamesPeters98:
Any chance you can fix the build @ludovicc? This would be really useful for me too

JamesPeters98:
Actually, I just realised I need this for Setters as well

Issue #983, opened 2022-06-22
spring-graphql compatible code generation


msl-at-fcb:
Now that spring-graphql is 1.0 and reasonably stable, it would be good to provide first class support for their structures.

Currently @QueryMapping, @MutationMapping and @SubscriptionMapping can be shoe-horned in via customAnnotationsMapping with something like:

```
  "customAnnotationsMapping": {
    "^Query\\.\\w+$": ["org.springframework.graphql.data.method.annotation.QueryMapping"],
    "^Mutation\\.\\w+$": ["org.springframework.graphql.data.method.annotation.MutationMapping"],
    "^Subscription\\.\\w+$": ["org.springframework.graphql.data.method.annotation.SubscriptionMapping"]
  }
```

But that falls short on parameterised resolvers, or resolvers for extensions which are generated separately which rely on spring's `@SchemaMapping(typeName = "MyGraphQLType")` (and so need to be manually added)

Likewise use of `@Argument` for all input parameters can be manually added to the implementing classes, but would be convenient to have on the generated interfaces.

There are other parameter types that can be provided (see https://docs.spring.io/spring-graphql/docs/current/reference/html/#controllers). Adding these with similar configuration as exists today for DataFetchingEnvironment would also be useful and would provide reasonably complete with what spring-graphql offers.

Also note that there is a feature request in spring-graphql for code generation: https://github.com/spring-projects/spring-graphql/issues/159 which would effectively be solved with this functionality.

JamesPeters98:
The custom annotations mapping you provided works well for me too.

The main issue for me is not being able to annotate the arguments with @Argument.

Maybe a customResolverArgumentAnnotations option would be good? It could just take a list of annotations to be applied to all arguments. But not the DataFetchingEnvironment etc

EndzeitBegins:
We recently adopted [Spring for GraphQL](https://docs.spring.io/spring-graphql/docs/current/reference/html/) and set up _graphql-java-codegen_ today.

Thanks to the comment of @msl-at-fcb we've got up and running relatively easy. 

In addition to using _customAnnotationsMapping_ for the root types `Query`, `Mutation` and `Subscription`, we also added `@SchemaMapping` for every field declared in `fieldsWithResolvers`.

```kotlin
tasks {
    named<GraphQLCodegenGradleTask>("graphqlCodegen") {
        // ...

        fieldsWithResolvers = setOf("Foo.bar")
[...]

PR #1005, opened 2022-09-19
Test for ResolverArgumentAnnotations on ParameterizedFieldsResolvers

velo:
Hi peeps,

While testing the new `ResolverArgumentAnnotations` I noticed they are not applied to parameterized resolvers.

I made a test to show it, and start my way on code to fix, but, I was not happy with how I got it working (boolean forceAnnotations = true)

I narrow it down to this call.
https://github.com/kobylynskyi/graphql-java-codegen/blob/master/src/main/java/com/kobylynskyi/graphql/codegen/mapper/FieldDefinitionsToResolverDataModelMapper.java#L239-L241

But not sure what I should do to fix it, should I add a boolean? Update `mappingContext`?

Any advice on how to do the fix is welcome

---

### Description

Related to #983 

---

Changes were made to:
- [ ] Codegen library - Java
- [ ] Codegen library - Scala
- [ ] Maven plugin
- [ ] Gradle plugin
- [ ] SBT plugin
` I

### DIFF
--- a/plugins/gradle/README.md
+++ b/plugins/gradle/README.md
@@ -17,7 +17,7 @@
 
 ```groovy
 plugins {
-  id "io.github.kobylynskyi.graphql.codegen" version "5.5.0"
+  id "io.github.kobylynskyi.graphql.codegen" version "5.6.0"
 }
 ```
 
@@ -31,7 +31,7 @@ buildscript {
     }
   }
   dependencies {
-    classpath "io.github.kobylynskyi.graphql.codegen:graphql-codegen-gradle-plugin:5.5.0"
+    classpath "io.github.kobylynskyi.graphql.codegen:graphql-codegen-gradle-plugin:5.6.0"
   }
 }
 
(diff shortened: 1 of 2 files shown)


PhilippS93:
can confirm, the Annotation is not added to parameterized resolvers

clement-buchart:
Here is my current workaround, to be applied here : https://github.com/kobylynskyi/graphql-java-codegen/blob/f488214ac4ffcf8cd7e44f79f6a02695482ae644/src/main/java/com/kobylynskyi/graphql/codegen/mapper/AnnotationsMapper.java#L125

```
        // 6. Add annotations for resolver arguments
        if (!Utils.isEmpty(mappingContext.getResolverArgumentAnnotations())
                && (mappingContext.getOperationsName().contains(parentTypeName) 
                   || (def instanceof InputValueDefinition) && !mappingContext.getDocument().getInputDefinitions().stream().map(o -> o.getName()).collect(Collectors.toList()).contains(parentTypeName) ) ) {
            annotations.addAll(mappingContext.getResolverArgumentAnnotations());
        }
```

Basically annotate anything that is a InputValue, but is not a InputType, leaving (i think) only argument resolvers.

@kobylynskyi Not sure if this is satisfactory or if there is an existing utils method to do this assertion ?

PR #1044, opened 2023-03-13
Add resolver argument annotation on parameterized field resolvers

clement-buchart:
---

### Description

Currently, the resolver argument annotation is not applied on parameterized field resolver, which is problematic when using this generator to apply the  org.springframework.graphql.data.method.annotation.Argument necessary to work with spring boot graphql integration.

This contains a commit that only adds a assertion file to demonstrate the issue, and another to fix it.

---

Changes were made to:
- [X] Codegen library - Java
- [X] Codegen library - Scala

### DIFF
--- a/src/main/java/com/kobylynskyi/graphql/codegen/mapper/AnnotationsMapper.java
+++ b/src/main/java/com/kobylynskyi/graphql/codegen/mapper/AnnotationsMapper.java
@@ -20,6 +20,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import static com.kobylynskyi.graphql.codegen.mapper.GraphQLTypeMapper.getDirectives;
 import static com.kobylynskyi.graphql.codegen.mapper.GraphQLTypeMapper.getMandatoryType;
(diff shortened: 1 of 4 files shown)


src/main/java/com/kobylynskyi/graphql/codegen/mapper/AnnotationsMapper.java
```diff
@@ -123,9 +124,16 @@ public List<String> getAnnotations(MappingContext mappingContext, String graphQL
             }
         }
         // 6. Add annotations for resolver arguments
-        if (!Utils.isEmpty(mappingContext.getResolverArgumentAnnotations())
-                && mappingContext.getOperationsName().contains(parentTypeName)) {
-            annotations.addAll(mappingContext.getResolverArgumentAnnotations());
+        if (!Utils.isEmpty(mappingContext.getResolverArgumentAnnotations())) {
+            if (mappingContext.getOperationsName().contains(parentTypeName)
+                || (def instanceof InputValueDefinition
+                && !mappingContext.getDocument().getInputDefinitions()
```
kobylynskyi, line 130: This method (`AnnotationsMapper.getAnnotations()`) is executed for **each and every GraphQL entity** (type/enum/interface/union/input/parameter/etc) which makes it an expensive method.

So collecting the names of all inputs should be done once in `MappingContext` and then should be reused here. (Similar caching approach was done for operation names / interface names / union names).

Please create a `Set` of input names in `MappingContext` and then here do `mappingContext.getInputsName().contains(parentTypeName)`.

kobylynskyi (changes requested):

src/main/java/com/kobylynskyi/graphql/codegen/mapper/AnnotationsMapper.java
clement-buchart: Thank you for the review. 
It is done.

kobylynskyi (approved):

Issue #1045, opened 2023-03-16
Upgrade com.graphql-java:graphql-java to resolve security vulnerabilities 


upendra-vedullapalli:
## Issue Description

*This repo is vulnerable to DoS attack([CVE-2022-37734](https://nvd.nist.gov/vuln/detail/CVE-2022-37734)) from `com.graphql-java:graphql-java:16.2`* 

## Steps to Reproduce

*Run a check on dependencies list using a plugin like [OWASP](https://github.com/jeremylong/DependencyCheck) to list all security vulnerabilities*

## Expected Result

*Dependency check report for vulnerabilities would list [CVE-2022-37734](https://nvd.nist.gov/vuln/detail/CVE-2022-37734)*

## Actual Result

*Dependency on `com.graphql-java:graphql-java` from version before 19.0 makes this repository vulnerable to DoS attack on parsing larger schemas*

## Your Environment and Setup

* **graphql-java-codegen version**: *5.6.0*
* **Build tool**: *Gradle*
* **Mapping Config**: *Default configuration*

kobylynskyi:
Thanks @upendrao for working on this. Your fix will be released in 5.7.0 very soon.

upendra-vedullapalli:
@kobylynskyi 
May I know when can we expect 5.7.0 released?

kobylynskyi:
I am waiting just for one PR to be merged for 5.7.0 and we should be good to go. Thanks.

PR #1046, opened 2023-03-16
fix: Updated dependencies to resolve security vulnerabilities

upendrao:
---

### Description

Related to #1045 

Updated following dependencies to resolve security vulnerabilities
- `com.graphql-java:graphql-java`
- `com.fasterxml.jackson.core:jackson-databind`
Updated unit tests to be able to parse test schemas with more tokens than the default size(15000) 

---

Changes were made to:
- [ ] Codegen library - Java
- [ ] Codegen library - Scala
- [ ] Maven plugin
- [ ] Gradle plugin
- [ ] SBT plugin
- [x] Dependencies

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -21,8 +21,8 @@ repositories {
 
 dependencies {
     compileOnly "org.freemarker:freemarker:2.3.31"
-    compileOnly "com.graphql-java:graphql-java:16.2"
-    compileOnly "com.fasterxml.jackson.core:jackson-databind:2.13.3"
+    compileOnly "com.graphql-java:graphql-java:20.0"
+    compileOnly "com.fasterxml.jackson.core:jackson-databind:2.14.2"
     compileOnly "com.typesafe:config:1.4.1"
 
     testImplementation "org.junit.jupiter:junit-jupiter-api:5.7.1"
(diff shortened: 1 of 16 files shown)


kobylynskyi (approved):

kobylynskyi:
Thanks @upendrao for the contribution!

upendrao:
@kobylynskyi I'm looking at the failure and shall submit a PR on that.

upendrao:
Hi @kobylynskyi ! Can I ask you a question here?
I have bunch of graphql [schemas](https://github.com/acryldata/datahub/tree/master/datahub-graphql-core/src/main/resources) from DataHub on top of which I want to write my own custom queries as we run from a GraphQL end point. How can I use this plugin to generate code for custom queries. 
For example I tried to add an additional graphql file in the schemas folder for the plugin but could not see any code generated for my custom query. Any leads?

upendrao:
Hi,
If I use 'extend type Query {' then I think I need to implement that query
on my own as it wont exist on GraphQL server.
For example I have a query as follows then I want to generate Java/Kotlin
code for that query so that I can easily deserialize the response into it.
query {
listDomains(
input: {}
) {
start
count
total
domains {
urn
properties {
name
}
}
}
}
The types defined in the base schema (ex: Domain here) have a mix of
nullable and non-nullable fields.
But in a query projection we (want to)list only a subset of the fields of a
specific type.
I am experiencing a challenge while deserializing such types when the
projection does not include a non-nullable field.
Is there any fix for this on the json codec?

Best Regards,
Upendra

Issue #146, opened 2020-05-15
Use lombok to reduce boilerplate code in generated model files


jyjyj:
Currently the model files generated are full of the typical OO-stuff: private variables and public getter/setters. Would it be helpful to generate less code by simply using lombok to annotate the generated classes?

kobylynskyi:
Hey @jyjyj 
I don't see any sense doing this because this boilerplate code is auto-generated.
Moreover, lombok annotations do not work for generated sources (I already tried 😄)

PR #1048, opened 2023-03-17
Made possible to use custom templates

velo:
---

### Description

Added the configuration for manually selecting templates.  This would allow anyone that needs/wants to modify a output, to do it by themselves.

For instance, I'm would like to use java record with lombok annotations, by using a custom template I can do it.

Related to #146 and potentially many others

---

Changes were made to:
- [x] Codegen library - Java
- [x] Codegen library - Scala
- [x] Maven plugin
- [x] Gradle plugin
- [x] SBT plugin

If the change seems reasonable, I will make the gradle plugin and SBT changes


src/main/java/com/kobylynskyi/graphql/codegen/generators/FreeMarkerTemplateFilesCreator.java
```diff
@@ -48,7 +48,13 @@ public static File create(MappingContext mappingContext,
         }
 
         try (FileWriter fileWriter = new FileWriter(javaSourceFile)) {
-            Template template = FreeMarkerTemplatesRegistry.getTemplateWithLang(language, templateType);
+            Template template;
+            if (mappingContext.getCustomTemplates().containsKey(templateType)) {
+              template = FreeMarkerTemplatesRegistry.getCustomTemplates(mappingContext.getCustomTemplates().get(templateType));
```
kobylynskyi, line 53: 1 - Don't make 2 map lookups (first `.containsKey()`, then `.get()`). Instead - do `.get()` and then check if the value is `null`.
2 - Use constant indentation of 4 spaces.

src/main/java/com/kobylynskyi/graphql/codegen/model/MappingConfig.java
```diff
@@ -1,5 +1,7 @@
 package com.kobylynskyi.graphql.codegen.model;
 
+import com.kobylynskyi.graphql.codegen.generators.FreeMarkerTemplateType;
+import java.util.EnumMap;
```
kobylynskyi, line 4: Unused import

src/main/java/com/kobylynskyi/graphql/codegen/model/MappingConfig.java
```diff
@@ -241,6 +245,28 @@ public void setCustomTypesMapping(Map<String, String> customTypesMapping) {
...
+     * Put custom template if absent.
+     *
+     * @param from the from
+     * @param to   the to
+     */
+    public void putCustomTemplatesIfAbsent(FreeMarkerTemplateType from, String to) {
+      if (customTemplates == null) {
+        customTemplates = new HashMap<>();
+      }
+      customTemplates.computeIfAbsent(from, k -> to);
```
kobylynskyi, line 258: Use 4 spaces indentation across all files

src/test/java/com/kobylynskyi/graphql/codegen/GraphQLCodegenAnnotationsTest.java
```diff
@@ -321,4 +322,14 @@ private void generate(String path) throws IOException {
                 TestUtils.getStaticGeneratedInfo(mappingConfig)).generate();
     }
 
+    @Test
+    void generate_CustomTemplates_Type() throws Exception {
```
kobylynskyi, line 326: This test case is not related to annotations, so please move it to a separate test class `GraphQLCodegenCustomTemplatesTest`

kobylynskyi:
Hey @velo 
Thanks for the contribution!
I like this feature.
Please address the comments I put in individual files. 
Also here are additional notes:
1 - Please fix the build, for some reason maven codegen fails with NPE: https://github.com/kobylynskyi/graphql-java-codegen/actions/runs/4445081094/jobs/7821033595?pr=1048
2 - If you can append a new record to the end of the configs description table in `codegen-options.md`.
3 - If you are familiar with Gradle/SBT plugins, please add support to those too.

Thank you!

src/main/java/com/kobylynskyi/graphql/codegen/model/MappingConfig.java
```diff
@@ -241,6 +245,28 @@ public void setCustomTypesMapping(Map<String, String> customTypesMapping) {
         this.customTypesMapping = customTypesMapping;
     }
 
+    /**
+     * Put custom template if absent.
```
kobylynskyi, line 249:
```suggestion
     * Provide a path to a custom template for the specific FreeMarker template type (if absent).
```

kobylynskyi (changes requested):

kobylynskyi (dismissed):

velo:
Fixed scalafmt

kobylynskyi (approved):

velo:
Found a small buh on this, hope I can patch for the release

PR #1052, opened 2023-03-20
Change customTemplates Map key to be type String

velo:
---

### Description

Related to #1048 

Only realized I was testing the wrong SNAPSHOT when I change the version from 5.6.1-SNAPSHOT to 5.7.0-SNAPSHOT.

Can't use the enum as key on the map, as build tools will inject strings.

Also did a minor improvement that allow to setup maven like this:

```
<configuration>
  <customTemplates>
    <TYPE>src/main/resource/type.ftl</TYPE>
  </customTemplates>
<configuration>
```

Before template need to be on classpath and was whole lot more complicated

---

Changes were made to:
- [X] Codegen library - Java
- [X] Codegen library - Scala
- [X] Maven plugin
- [X] Gradle plugin
- [X] SBT plugin


kobylynskyi:
@velo changes look good. Can you please fix the build?

velo:
> @velo changes look good. Can you please fix the build? (checkstyle issues)

BTW, can I run checkstyle locally?

I enabled the plugin, but seems I don't get the same warnings...

kobylynskyi (approved):

kobylynskyi:
@velo yes, you can. There are plugins for different IDEs and you can import the rules from the following folder: [config/check-style](https://github.com/kobylynskyi/graphql-java-codegen/tree/master/config/checkstyle)

PR #1059, opened 2023-03-23
ci: add dependabot configuration

yeikel:
### Description

See https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/

---

Changes were made to:

This is another type of change not directly related to any ecosystem


kobylynskyi:
@yeikel will this include dependencies from all subprojects?
There are following projects inside this repo:
* build.gradle
* plugins/gradle/graphql-java-codegen-gradle-plugin/build.gradle
* plugins/gradle/example-client/build.gradle
* plugins/gradle/example-server/build.gradle
* plugins/gradle/example-client-kotlin/build.gradle
* plugins/maven/graphql-java-codegen-maven-plugin/pom.xml
* plugins/maven/example-client/pom.xml
* plugins/maven/example-server/pom.xml

yeikel:
> 

Are these projects fully independent (ie: not connected to the main build.gradle file)

If so, then we need to configure them explicitly

yeikel:
@kobylynskyi You can see it in action in my fork : https://github.com/yeikel/graphql-java-codegen/pulls

kobylynskyi:
@yeikel thanks for adding dependabot config back. We had it previously for a very long time and then I disabled it because of too many PRs for `example` projects. So I suggest setting it to `monthly` and keep `weekly` only for 2 projects (main, gradle plugin and maven plugin). Thanks.

yeikel:
> @yeikel thanks for adding dependabot config back. We had it previously for a very long time and then I disabled it because of too many PRs for `example` projects. So I suggest setting it to `monthly` and keep `weekly` only for 2 projects (main, gradle plugin and maven plugin). Thanks.

One option we can explore is to limit the number of open pull requests for certain ecosystems. That should reduce the number of pull requests for the examples 

See https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#open-pull-requests-limit

kobylynskyi:
> One option we can explore is to limit the number of open pull requests for certain ecosystems. That should reduce the number of pull requests for the examples
See https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#open-pull-requests-limit
If that works for you, can still leave others enabled but maybe set open-pull-requests-limit to 1 for the examples to still have the benefit but reduce the noise.

Based on the amount of dependabot PRs I had previously for example projects, I am keen towards settings the interval to monthly (without a limit on the amount)

yeikel:
@kobylynskyi Thank you for merging it 

You should consider creating a CODEOWNERS file and making it mandatory for all reviews so that the reviewer configuration is not needed here

kobylynskyi:
Thanks @yeikel for creating this PR!

yeikel:
Note that due to how dependabot works, this file needs to land in `master` (the default branch) to work

Issue #1151, opened 2023-04-07
Usage of `Task.project` prevents use of configuration cache


bmuschko:
## Issue Description

Run build with `--configuration-cache`. You will receive the following error message for your plugin.

> invocation of `Task.project` at execution time is unsupported.

## Steps to Reproduce

Apply your plugin. Run build with `--configuration-cache`.

## Expected Result

Gradle does not report configuration cache error anymore.

## Actual Result

Error message is rendered.

## Your Environment and Setup

* **graphql-java-codegen version**: *[v5.7.1](https://github.com/kobylynskyi/graphql-java-codegen/releases/tag/v5.7.1)*
* **Build tool**: *Gradle 8.0*

bmuschko:
Is this something you are planning to fix soon or would you rather want to receive a pull request?

kobylynskyi:
Yes, you are right. I've removed it from 5.7.2 release.

esfomeado:
@kobylynskyi I gave it a try but I wasn't able to get it to work.

Are you able to look at this?

Thanks

vvvlasov:
Hi! This is still an issue as of 5.10.0, do you need any help to get it done?

esfomeado:
@vvvlasov Yes please

xenoterracide:
yep, went to look/report today and this blocks me from using this plugin. I guess I'll go about this the otherway code -> schema instead of schema -> code

PR #1157, opened 2023-04-14
Removed use of project to enable configuration cache

esfomeado:
---

### Description

Related to #1151

---

Changes were made to:
- [ ] Codegen library - Java
- [ ] Codegen library - Scala
- [ ] Maven plugin
- [X] Gradle plugin
- [ ] SBT plugin


esfomeado:
It's not as straightforward as I thought.

Will investigate further

bmuschko:
Thanks for working on this. To determine the compatibility with the configuration cache, you may want to set up a functional test with [TestKit](https://docs.gradle.org/current/userguide/test_kit.html) that applies your plugin and executes a build with using `--configuration-cache`. It'll be easier than building the binary and trying it out on a sample project.

Issue #1375, opened 2023-11-06
"...on" is skipped in all$ methods for union and interface projections


sdaurea:
## Issue Description

Currently the generated method `all$` for projections does not include fields that start with "on" prefix.

## Steps to Reproduce

It is already reproduced by the test [here](https://github.com/kobylynskyi/graphql-java-codegen/blob/b13e0591cd2831e1eb9cfd2d30751ed97e312a1e/src/test/resources/expected-classes/response/LocationResponseProjection.java.txt#L39). The same can be reproduced for the union, but there is no such test for the union.

## Expected Result

It is expected that `onCustomerLocation` and `onDealerLocation` will be included into the `all$` method.

## Actual Result

Only `id`, `locationType` and `typename` are included.

## Your Environment and Setup

* **graphql-java-codegen version**: *5.8.0*
* **Build tool**: *Gradle*

jxnu-liguobin:
Thank you for discovering it. This is a known legacy work!

kobylynskyi:
The PR was merged. Will be released in 5.9.0. Stay tuned.

PR #1376, opened 2023-11-06
Do not skip "...on" in all$ methods for union and interface projections

sdaurea:
---

### Description

Related to #1375

---

Changes were made to:
- [x] Codegen library - Java
- [x] Codegen library - Scala
- [ ] Maven plugin
- [ ] Gradle plugin
- [ ] SBT plugin

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -9,7 +9,7 @@ plugins {
     id "org.sonarqube" version "4.0.0.2929"
 }
 
-def graphqlCodegenVersion = '5.8.1-SNAPSHOT' // This variable used in the automatic release process
+def graphqlCodegenVersion = '5.8.0' // This variable used in the automatic release process
 
 group = "io.github.kobylynskyi"
 version = graphqlCodegenVersion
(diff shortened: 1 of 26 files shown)


jxnu-liguobin:
When you open it, does it mean you are in need of it? This is great. Can you test if it really meets the requirements? Thank you!

Because I skipped it directly to avoid testing the union.

alaa-nassef:
> When you open it, does it mean you are in need of it? This is great. Can you test if it really meets the requirements? (Test its results at runtime ) Thank you!
> 
> Because I skipped it directly to avoid testing the union.

@jxnu-liguobin to answer your first question, yes, we do need this in the project we're working on and it is currently kind of a blocker for us. As for testing results at runtime, we were planning to do this, but we have to publish our own version first and use it instead, which would need some more time

kobylynskyi:
@alaa-nassef looks good!. Thanks for the contribution!

derwild:
I'm not sure about your release schedule, but can you already say, when the 5.9.0 maven plugin will be released?

kobylynskyi:
@derwild targeting this week.

PR #1377, opened 2023-11-06
Added an option to render Kotlin functions with suspend modifier

thevietto:
---

### Description

Added a configuration option for Kotlin generator to render functions with `suspend` modifier:

when disabled:
```
fun sayHello(): String
```
when enabled:
```
suspend fun sayHello(): String
```

I know this could be solved by providing a custom Freemarker template to the config, but it seems like supporting Kotlin coroutines is something one would expect to get out of the box.

---

Changes were made to:
- [ ] Codegen library - Java
- [ ] Codegen library - Scala
- [x] Maven plugin
- [x] Gradle plugin
- [ ] SBT plugin


jxnu-liguobin:
It looks good, looking forward to you completing it. By the way. SBT cannot compile kotlin by default unless using third-party plugins. So you don't need to consider SBT. And I am considering removing the code related to kotlin from the sbt plugin in this library.

thevietto:
> It looks good, looking forward to you completing it. By the way. SBT cannot compile kotlin by default unless using third-party plugins. So you don't need to consider SBT. And I am considering removing the code related to kotlin from the sbt plugin in this library.

thanks! Yeah I didn't touch the SBT module
codestyle is fixed, should be good now

jxnu-liguobin (approved):

kobylynskyi (approved):

kobylynskyi:
Looks good! Thanks for the contribution.

Issue #966, opened 2022-03-08
Support for Optional argument handling


PhilippS93:
First of all, thank you for the library, it's very great! 👏

**Is your feature request related to a problem? Please describe.**
GraphQL supports 3 states of input arguments. 
- They can be set but null
- They can be not set
- They can be set and not null

Java only represents two states, set or not set. But the information if the property was present in the request is missing.
GraphQL Frameworks for java like https://github.com/graphql-java-kickstart/graphql-java-tools/pull/404 support this feature by using Optional<>-java class.

**Describe the solution you'd like**
Add a toggle in the generation to generate Optional<> fields in input-classes.

e.g. for

```graphql
input EditUserInput {
    id: ID!
    name: String!
    address: String
}
```
it currently generates:

```java
public class EditUserInputDTO implements java.io.Serializable {

    @javax.validation.constraints.NotNull
    private String id;
    @javax.validation.constraints.NotNull
    private String name;
    private String address;
  //...
}
```
but now, we could generate:

```java
public class EditUserInput DTO implements java.io.Serializable {

    @javax.validation.constraints.NotNull
    private String id;
    @javax.validation.constraints.NotNull
    private String name;
    private Optional<String> address;
  //...
}
```

This would allow us to check if the property should be deleted (Optional.empty() == false but null) or was just not set by the frontend (Optional.empty() == true) and must not be considered for updating or deletion.

**Describe alternatives you've considered**
None

kobylynskyi:
Java 8's Optional was mainly intended for return values from methods, and not for properties of Java classes, as described in [Optional in Java SE 8](http://blog.joda.org/2014/11/optional-in-java-se-8.html):

> Of course, people will do what they want. But we did have a clear intention when adding this feature, and it was not to be a general purpose Maybe or Some type, as much as many people would have liked us to do so. Our intention was to provide a limited mechanism for library method return types where there needed to be a clear way to represent "no result", and using null for such was overwhelmingly likely to cause errors.
> 
> The key here is the focus on use as a return type. The class is definitively not intended for use as a property of a Java Bean. Witness to this is that Optional does not implement Serializable, which is generally necessary for widespread use as a property of an object.

So I am parking this feature request for now.

PhilippS93:
Thanks for your response. 
Because graphql-java-tools already support this feature, how could we solve this by using the codegen lib?

MartyWind:
[...]

PR #1450, opened 2024-01-16
Support undefined input parameter

rwo-trackunit:
---
### Description
When updating an object we want to support patch like procedure:
- If value is not send, do not change the value
- If value is null set the value to null
- If value is set, set that value

In #966 it was debated if we could us Optional to solve this, ultimately agree that it would not be appropriate. 
> Java 8's Optional was mainly intended for return values from methods, and not for properties of Java classes, as described in [Optional in Java SE 8](http://blog.joda.org/2014/11/optional-in-java-se-8.html)

This PR tries to solve it by introducing a new wrapper for input value that is not mandatory and is not a list.

#### Example:
```
input BoundingBoxInput {
    nw: CoordinatesInput!
    se: CoordinatesInput!
    zoom: Float
    name: String = null
    description: String = "a very good desc"
}
```
becomes
```
public class BoundingBoxInput implements java.io.Serializable {
    private CoordinatesInput nw;
    private CoordinatesInput se;
    private GraphQLInputParameter<Double> zoom = GraphQLInputParameter.undefined();
    private GraphQLInputParameter<String> name = GraphQLInputParameter.withNull();
    private GraphQLInputParameter<String> description = GraphQLInputParameter.withValue("a very good desc");
    
    ...
```

For convenience GraphQLInputParameter offers:
- `isDefined()`
- `isUndefined()`
- `getValue` (with throws)
- `getValueOrDefault(...)`

This is all locked behind the configuration `useWrapperForNullableInputTypes`.

#### TODO:
As I have no experience with Kotlin or Scala, this currently only works for Java.

---

Changes were made to:
- [X] Codegen library - Java
- [X] Codegen library - Scala


src/main/java/com/kobylynskyi/graphql/codegen/model/MappingContext.java
```diff
@@ -259,6 +259,11 @@ public Boolean getUseOptionalForNullableReturnTypes() {
         return config.getUseOptionalForNullableReturnTypes();
     }
 
+    @Override
```
jxnu-liguobin, line 262: we should also add it to sbt. This is easy to write

rwo-trackunit:
I could also use some suggestions on naming: `undefined` small a bit like Javascript.

jxnu-liguobin:
> I could also use some suggestions on naming: `undefined` smells a bit like Javascript.

Maybe we don't need it, only isDefined is needed.

rwo-trackunit:
> > I could also use some suggestions on naming: `undefined` smells a bit like Javascript.
> 
> Maybe we don't need it, only isDefined is needed.

Yea, I should only have one of Defined/Undefined. Maybe we should adopt the `Optional.isPresent()` naming. Although I'm keen to show this is not just a homemade Optional.
Why is naming always the hardest part of coding?

src/main/java/com/kobylynskyi/graphql/codegen/model/MappingContext.java
rwo-trackunit: Can you give me a hint on where/how that would be done? A bit out of my depth, but happy to give it a try.

jxnu-liguobin:
> > > I could also use some suggestions on naming: `undefined` smells a bit like Javascript.
> > 
> > 
> > Maybe we don't need it, only isDefined is needed.
> 
> Yea, I should only have one of Defined/Undefined. Maybe we should adopt the `Optional.isPresent()` naming. And `.get` and `.orElse` Although I'm keen to show this is not just a homemade Optional. Why is naming always the hardest part of coding?

People from different languages and backgrounds have different norms. 😄

src/main/java/com/kobylynskyi/graphql/codegen/model/MappingContext.java
rwo-trackunit: I tried adding it: [Commit](https://github.com/kobylynskyi/graphql-java-codegen/pull/1450/commits/da664558f25925bdbdb65262912bd4cb259d073f)
Let me know if this covers all changes needed for SBT.

jxnu-liguobin:
It looks good, just describe the option in the markdown document and wait for @kobylynskyi review.

rwo-trackunit:
After some feedback, this has been rewritten to use [ArgumentValue](https://docs.spring.io/spring-graphql/docs/current/api/org/springframework/graphql/data/ArgumentValue.html), rather than generating its own class that solves the same problem.

AllirionX:
Thanks for the PR. If projects are not based on spring-graphql, it might be valuable to let the wrapper type up to the user, like this:
`<useWrapperForNullableInputTypes>org.springframework.graphql.data.ArgumentValue</useWrapperForNullableInputTypes>`
`<useWrapperForNullableInputTypes>java.util.Optional</useWrapperForNullableInputTypes>`
`<useWrapperForNullableInputTypes>custom.example.Undefined</useWrapperForNullableInputTypes>`

rwo-trackunit:
> Thanks for the PR. If projects are not based on spring-graphql, it might be valuable to let the wrapper type up to the user, like this: `<useWrapperForNullableInputTypes>org.springframework.graphql.data.ArgumentValue</useWrapperForNullableInputTypes>` `<useWrapperForNullableInputTypes>java.util.Optional</useWrapperForNullableInputTypes>` `<useWrapperForNullableInputTypes>custom.example.Undefined</useWrapperForNullableInputTypes>`

I thought about that, and even have had it compatible with Optional at an earlier point in the PoC.
But to support default values, each wrapper type requires 3 string replacement strings to support fx. `ArgumentValue.omitted()` vs `Optional.empty()`. Adding that in config quickly becomes very messy.

Alternative solution would be an interface, but `ArgumentValue` doesn't extend one.

Last good idea is a factory, that is default provided supporting `ArgumentValue`, but could be replaced with a custom factory. 
That could work in general, but could never work with `spring-graphql`, since [GraphQlArgumentBinder](https://github.com/spring-projects/spring-graphql/blob/e054fe13cf92b5b3bc2e3a56df1f8080989170cc/spring-graphql/src/main/java/org/springframework/graphql/data/GraphQlArgumentBinder.java#L181) only supports `Optional` and `ArgumentValue` wrappers. 

Other alternative solution would be a custom object provided by this project, but that doesn't support `spring-graphql` for the same reason.

If there is further request for some solution to support more wrappers (be it the factory or some other good idea), I'm happy to code it up. But I feel this PR is big enough.

kobylynskyi (approved):
@rwo-trackunit good job! Your changes look good. I will merge this soon, wait for some other PRs to be merged and release it ASAP.
Thanks!

rwo-trackunit:
Once merged, don't hesitate to tag me in any found bugs, or suggested changes to this corner of the project.

rwo-trackunit:
Consider squashing this PR. The commits in here includes several PoCs, so it may give a confusing git history.

kobylynskyi:
This was released in 5.10.0. Thanks!

